PR #21, opened 2017-06-05
Write Arrow buffer into host shared memory in CPU mode

trxcllnt:
cc: @andrewseidl 

This PR is a starting point for writing the serialized Arrow buffer to host shared memory in CPU mode via [`shmget`](http://man7.org/linux/man-pages/man2/shmget.2.html). I've tested it works locally on macOS Sierra 10.12.5, though it should work on any System V platform.

My C++ fu is not strong, so I cargo-culted the ThriftHandler. I implemented new code paths to reduce the risk of introducing new bugs. I expect someone from the team will want to delete or rewrite this code to conform to the coding guidelines.

The core logic is in [`ResultSet::getArrowCopy`](https://github.com/mapd/mapd-core/compare/master...graphistry:shm-ipc#diff-3055c0b6e6034b731c9cdad3b3cd3121R529). I've added comments inline with the following questions/todos:
- I've used stdlib rand() to [generate unique shmid keys](https://github.com/mapd/mapd-core/compare/master...graphistry:shm-ipc#diff-3055c0b6e6034b731c9cdad3b3cd3121R541). Are there any flaws in this approach?
- if [`shmget` or `shmat` fail](https://github.com/mapd/mapd-core/compare/master...graphistry:shm-ipc#diff-3055c0b6e6034b731c9cdad3b3cd3121R560), the error should be surfaced to the caller (needs rewrite following mapd-core guidelines)
- it should be possible to serialize the arrow buffer directly to shared memory, but I'm not familiar with the C++ Streams API. I've [hacked it with a memcpy](https://github.com/mapd/mapd-core/compare/master...graphistry:shm-ipc#diff-3055c0b6e6034b731c9cdad3b3cd3121R574) for now

Here's an example of calling the `sql_execute_df` method from node (see more in our [graphistry/rxjs-mapd](https://github.com/graphistry/rxjs-mapd) repo):

```javascript
// Assuming mapd-core is running with default settings and flights_2008_10k test database
import * as shm from 'shm-typed-array';
import { createClient, createConnection, TBinaryProtocol, TBufferedTransport } from 'thrift';
// These libs aren't in npm yet
// Arrow's JS impl has a bug that prevents it from reading data from
// columns that don't contain nulls. I patched it locally, and will PR them soon
import { getReader } from 'apache/arrow';
// run `thrift --gen js:node mapd.thrift` to get this code
import { Client as MapDClient } from 'mapd-thrift';

const client = createClient(MapDClient,  createConnection(`localhost`, 9091, {
  protocol: TBinaryProtocol, transport: TBufferedTransport
}));

client.connect(`mapd`, `HyperInteractive`, `mapd`)
  .then((session) => client.sql_execute_df(session, `SELECT origin_lat, origin_lon FROM flights_2008_10k WHERE dest_city ILIKE 'dallas'`, -1))
  .then(({ df_handle }) => getReader(shm.get(+df_handle, 'Buffer')))
  .then((arrow) => console.log(`col names: ${arrow
      .getSchema()
      .map(({ name }) => name)
      .join(', ')
    }`)
  );
```

### DIFF
--- a/QueryEngine/ResultSetConversion.cpp
+++ b/QueryEngine/ResultSetConversion.cpp
@@ -18,6 +18,14 @@
 #include "Execute.h"
 
 #ifdef ENABLE_ARROW_CONVERTER
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/shm.h>
+#include <errno.h>
+#include <string>
+
 #include "arrow/builder.h"
 #include "arrow/buffer.h"
 #include "arrow/io/memory.h"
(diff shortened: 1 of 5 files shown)


m1mc:
Hi @trxcllnt, thanks for your commits, we will review and merge them soon.

ThriftHandler/MapDHandler.cpp
```diff
@@ -641,6 +641,62 @@ void MapDHandler::sql_execute(TQueryResult& _return,
...
+                                 const int32_t first_n) {
+#ifdef ENABLE_ARROW_CONVERTER
+  const auto session_info = MapDHandler::get_session(session);
+  int64_t execution_time_ms = 0;
+  LOG(INFO) << query_str;
+  int64_t total_time_ms = measure<>::execution([&]() {
+    SQLParser parser;
+    std::list<std::unique_ptr<Parser::Stmt>> parse_trees;
+    std::string last_parsed;
+#ifdef HAVE_CALCITE
```
asuhan: It's safe to assume that `HAVE_CALCITE` and `HAVE_RAVM` are always defined at this point. We should find some time to remove the legacy clutter, sorry about it.

QueryEngine/ResultSetConversion.cpp
```diff
@@ -518,6 +526,65 @@ arrow::RecordBatch ResultSet::convertToArrow(const std::vector<std::string>& col
...
+  // copy the arrow records buffer to shared memory
+  // TODO(ptaylor): I'm sure it's possible to tell Arrow's StreamWriter to write
+  // directly to the shared memory segment as a sink, but my c++ fu is not strong,
+  // so memcpy'ing for now.
+  memcpy(ipc_ptr, serialized_records->data(), serialized_records->size());
+  // detach from the shared memory segment
+  shmdt(ipc_ptr);
+  // cast the shared memory key to a string, and then to a char vec, so we can
+  // keep the same method signature as the original `getArrowDeviceCopy` below.
+  std::string handle_str = std::to_string(key);
```
m1mc: @trxcllnt is any reason why you don't serialize the key into 2-char vector like GPU handle serialization?

QueryEngine/ResultSetConversion.cpp
```diff
@@ -518,6 +526,65 @@ arrow::RecordBatch ResultSet::convertToArrow(const std::vector<std::string>& col
...
+  if ((ipc_ptr = (char*)shmat(shmid, NULL, 0)) == (char *) -1) {
+      // TODO(ptaylor): Surface this error gracefully instead of crashing
+      perror("shmat");
+      exit(1);
+  }
+  // copy the arrow records buffer to shared memory
+  // TODO(ptaylor): I'm sure it's possible to tell Arrow's StreamWriter to write
+  // directly to the shared memory segment as a sink, but my c++ fu is not strong,
+  // so memcpy'ing for now.
+  memcpy(ipc_ptr, serialized_records->data(), serialized_records->size());
```
m1mc: @trxcllnt you probably want to use MemoryMappedFile as the sink.

QueryEngine/ResultSetConversion.cpp
```diff
@@ -518,6 +526,65 @@ arrow::RecordBatch ResultSet::convertToArrow(const std::vector<std::string>& col
...
+      perror("shmget");
+      exit(1);
+    }
+    key = rand();
+  }
+  // get a pointer to the shared memory segment
+  char *ipc_ptr;
+  if ((ipc_ptr = (char*)shmat(shmid, NULL, 0)) == (char *) -1) {
+      // TODO(ptaylor): Surface this error gracefully instead of crashing
+      perror("shmat");
```
m1mc: @trxcllnt we better throw an exception when running into unexpected error.

QueryEngine/ResultSetConversion.cpp
```diff
@@ -518,6 +526,65 @@ arrow::RecordBatch ResultSet::convertToArrow(const std::vector<std::string>& col
   return arrow::RecordBatch(schema, row_count, columns);
 }
 
+// WARN(ptaylor): users are responsible for detaching and removing shared memory segments, e.g.,
+//   int shmid = shmget(...);
+//   char *ipc_ptr = (char* shmat(shmid, ...));
+//   ...
+//   shmdt(ipc_ptr);
+//   shmctl(shmid, IPC_RMID, 0);
+std::tuple<std::shared_ptr<arrow::Buffer>, std::vector<char>, int64_t> ResultSet::getArrowCopy(
```
asuhan: Maybe define a structure instead of using `std::tuple` to package the results. I guess it's fine since you immediately destructure with `std::tie` at the call site anyway.

QueryEngine/ResultSetConversion.cpp
trxcllnt: Nope! Just limited in my knowledge, really I'm like a baby here.

QueryEngine/ResultSetConversion.cpp
trxcllnt: 100% agree, just not acquainted with the error handling idioms of the project, so hoping you guys can provide guidance here (also line 560)

QueryEngine/ResultSetConversion.cpp
trxcllnt: Works for me. I tried to keep the method signature the same as `getArrowDeviceCopy` for consistency, but I'm happy to defer to the team's preferences here.

QueryEngine/ResultSetConversion.cpp
```diff
@@ -518,6 +526,65 @@ arrow::RecordBatch ResultSet::convertToArrow(const std::vector<std::string>& col
...
+  // incremental keygen would be deterministically worst-case. If we use a hash
+  // (like djb2) + nonce, we could still get collisions if multiple clients specify
+  // the same nonce, so using rand() in lieu of a better approach
+  // TODO(ptaylor): Is this common? Are these assumptions true?
+  key_t key = rand();
+  int shmsz = (int) serialized_records->size();
+  int shmid;
+  // IPC_CREAT - indicates we want to create a new segment for this key if it doesn't exist
+  // IPC_EXCL - ensures failure if a segment already exists for this key
+  while ((shmid = shmget(key, shmsz, IPC_CREAT | IPC_EXCL | 0666)) < 0) {
```
trxcllnt: I'm no expert in IPC, but from what @lmeyerov and I could tell, it seemed a file descriptor based solution could potentially run up against system limits on the number of simultaneous open FDs per process. Additionally, the node lib for attaching to shared memory segments is more active, and seems to reliably `shmctrl` on GC, even if the process crashes. Though I'm happy to defer to the expertise of the team here.

QueryEngine/ResultSetConversion.cpp
m1mc: @trxcllnt you can leave this change to return type to me, we are coming up with new thrift API to merge CPU and GPU path.

asuhan:
Thanks! We'll keep refining this in master.

PR #41, opened 2017-07-06
add --no-frontend option to startmapd script

trxcllnt:
adds a `--no-frontend` option that doesn't start the `mapd_web_server`, so the script can be used for starting just mapd + calcite. For example:

```
$ ../startmapd --no-frontend --non-interactive --data /tmp/databases/MyMapdDB --allow-loop-joins --enable-watchdog=false --enable-dynamic-watchdog=false
```


dwayneberry:
If you just want to run `mapd_server` and not `mapd_web_server` why not just run `mapd_web` server direct.  The only reason `startmapd` exists really is to simplify starting both servers in test env?

trxcllnt:
@dwayneberry this has been the most reliable way to ensure mapd and calcite both exit. I've had issues with one or the other dangling after segfault or sigkill. This is especially helpful for nodejs integration tests, in which the test runner starts mapd pointing to test databases dir.

andrewseidl:
Alternatively we could change the default to not open the browser, instead have a `--open` option if you really want a browser to pop up.

@tmostak thoughts?

dwayneberry:
@trxcllnt the issue with the restart after failure should be resolved with some recent changes to repo.  All orphaned servers should e cleaned with start of next server.  Are you running latest build.
If you are still seeing issues after segfaults please let us know reproduction steps so we can resolve.

trxcllnt:
for integration tests and migration scripts (we're getting ready for a big data migration from our binary format to mapd), we really don't want the mapd_web_server to start at all. I don't mind having a custom startup script, but figured other people might have the same needs. Adding an option here seemed like a relatively uncontroversial and backwards-compatible change to make this happen. Obvs it's up to you guys! 👍

dwayneberry:
@trxcllnt  Just to be super clear.  We do not recommend `startmapd` as a production way to start and stop the MapD database.  We would recommend `systemd` as the tool to management a real server deployment.

trxcllnt:
@dwayneberry I'm running the latest as of ten minutes ago on OS X 10.12.5, and sig-killing the server still hangs. I'm doing `./bin/mapd_server` followed by `ctrl-c` after it's up. Same issue if I start/kill the server in node.

trxcllnt:
> Just to be super clear. We do not recommend startmapd as a production way to start and stop the MapD database. We would recommend systemd as the tool to management a real server deployment.

Totally yeah, I'm just doing local dev and prototyping. Our prod environment will have everything containerized.

dwayneberry:
@trxcllnt Oh your running your MapD server on OS X?
All bets are off then ;-)  i'll leave this conversation alone now.

dwayneberry:
@trxcllnt I can reproduce your issue on Linux as well.  i am looking into it.  I'll never be mean to OS X again :-).

dwayneberry:
@trxcllnt i have hopefully stopped the issue you were seeing in ctrl-c shutdown   using the `mapd_server` directly with d5701bdd2e9e6.  if you could double check it works in your OS X env that would be much appreciated.

trxcllnt:
@dwayneberry yep, works great on my side.

dwayneberry:
@trxcllnt Thanks for helping test

dwayneberry:
I think we can close this one as all seems to be working as expected and no need to add changes

PR #46, opened 2017-07-13
JDBC isReadOnly return false

kkraus14:
Connection allows writes via JDBC by default and setReadOnly is not yet implemented. Function should be able to return false for compatibility without issue.

### DIFF
--- a/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDConnection.java
+++ b/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDConnection.java
@@ -205,12 +205,7 @@ public class MapDConnection implements java.sql.Connection {
     // never should get here
     return true;
   }
-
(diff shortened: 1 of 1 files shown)


dwayneberry:
We should check on the connection whether it really is read only or not.

The info is available with API `get_server_status`

Code should be more like this

```
@Override
  public boolean isReadOnly() throws SQLException { //logger.debug("Entered");
    try {
      logger.debug("Session at close is " + session);
      if (session != null) {
        TServerStatus server_status = client.get_server_status(session);
        return server_status.read_only;
      }
    } catch (TMapDException ex) {
      throw new SQLException("get_server_status failed." + ex.toString());
    } catch (TException ex) {
      throw new SQLException("get_server_status failed." + ex.toString());
    }
    // never should get here
    return true;
  }
```

PR #65, opened 2017-09-03
Migrate Arrow-related code to incorporate IPC code improvements since 0.4.x, simplifications and perf improvements

wesm:
* Use up-to-date Arrow IPC serialization / read APIs
* Use vector appends with Arrow builders (cf ARROW-1468)
* Avoid extra full result set memory copy in multithreaded fetches
* Use library calls for pretty printing (cf ARROW-1396)

### DIFF
--- a/QueryEngine/ArrowResultSet.cpp
+++ b/QueryEngine/ArrowResultSet.cpp
@@ -16,11 +16,12 @@
 
 #ifdef ENABLE_ARROW_CONVERTER
 #include "ArrowResultSet.h"
+#include "ArrowUtil.h"
 #include "RelAlgExecutionDescriptor.h"
 
-#include <arrow/array.h>
+#include <arrow/api.h>
 #include <arrow/io/memory.h>
-#include <arrow/ipc/reader.h>
+#include <arrow/ipc/api.h>
 
 namespace {
 
@@ -47,13 +48,14 @@ SQLTypeInfo type_from_arrow_field(const arrow::Field& field) {
 
 }  // namespace
 
-ArrowResultSet::ArrowResultSet(const std::shared_ptr<arrow::Schema>& schema,
-                               const std::shared_ptr<arrow::RecordBatch>& record_batch,
-                               const std::vector<std::shared_ptr<arrow::PoolBuffer>>& pool_buffers)
-    : record_batch_(record_batch), pool_buffers_(pool_buffers), crt_row_idx_(0) {
-  CHECK_EQ(schema->num_fields(), record_batch->num_columns());
+ArrowResultSet::ArrowResultSet(const std::shared_ptr<arrow::RecordBatch>& record_batch)
+    : record_batch_(record_batch), crt_row_idx_(0) {
+  auto schema = record_batch->schema();
   for (int i = 0; i < schema->num_fields(); ++i) {
-    column_metainfo_.emplace_back(schema->field(i)->name(), type_from_arrow_field(*schema->field(i)));
+    std::shared_ptr<arrow::Field> field = schema->field(i);
+    SQLTypeInfo type_info = type_from_arrow_field(*schema->field(i));
+    column_metainfo_.emplace_back(field->name(), type_info);
+    columns_.emplace_back(std::move(record_batch->column(i)));
   }
 }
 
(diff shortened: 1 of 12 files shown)


asuhan:
Thanks, the API simplifications look very nice. Regarding exceptions, it all depends on their nature. If it's running out of memory for example, you can throw `OutOfMemory` defined in `DataMgr/BufferMgr/BufferMgr.h`.

wesm:
OK, I'll add a function to create different exception types based on the type of error status

QueryEngine/ArrowResultSet.cpp
```diff
@@ -68,36 +69,36 @@ std::vector<TargetValue> ArrowResultSet::getNextRow(const bool translate_strings
     const auto& column_typeinfo = getColType(i);
     switch (column_typeinfo.get_type()) {
       case kSMALLINT: {
-        CHECK(dynamic_cast<const arrow::NumericArray<arrow::Int16Type>*>(&column));
-        const auto& i16_column = static_cast<const arrow::NumericArray<arrow::Int16Type>&>(column);
```
wesm: I do not recommend using `arrow::NumericArray` -- use the typedefs

QueryEngine/ResultSetConversion.cpp
```diff
@@ -206,299 +188,136 @@ TypePtr get_arrow_type(const SQLTypeInfo& mapd_type, const std::shared_ptr<Array
...
+    auto typed_builder = static_cast<BuilderType*>(this->raw_builder);
+
+    int64_t length = static_cast<int64_t>(vals.size());
+
+    if (this->field->nullable()) {
+      CHECK(is_valid.get());
+      ARROW_THROW_NOT_OK(typed_builder->Append(vals.data(), length, *is_valid));
+    } else {
+      ARROW_THROW_NOT_OK(typed_builder->Append(vals.data(), length));
     }
```
wesm: Vector appends

asuhan:
@wesm Thanks, I'll have a look tomorrow. For testing, have a look at https://github.com/mapd/mapd-core/blob/master/run_sanity_tests. You can run it as is, but you only need to run `ExecuteTest` (keep the first 9 lines unchanged).

wesm:
Yep, I found that. I'll step around gdb and fix the bugs I introduced, probably tomorrow later in the day

QueryEngine/ArrowUtil.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "arrow/status.h"
+#include "arrow/util/macros.h"
+
+void arrow_status_throw(const ::arrow::Status& s);
+
+#define ARROW_THROW_NOT_OK(s)                   \
+  do {                                          \
+    ::arrow::Status _s = (s);                   \
+    if (ARROW_PREDICT_FALSE(!_s.ok())) {        \
```
asuhan: Alternatively, you can use `UNLIKELY` from `Shared/likely.h`. I guess everyone has a version of this macro somewhere in the codebase.

QueryEngine/ArrowResultSet.cpp
asuhan: Nice, I didn't notice them when I wrote this.

asuhan:
Do the tests pass now?

wesm:
Not yet, I am going to look at them hopefully before I go to bed tonight

wesm:
I think it would be best to defer improvements re: direct writes to GPU device memory (i.e. https://github.com/apache/arrow/blob/master/cpp/src/arrow/gpu/cuda_arrow_ipc.h#L50) to a subsequent patch.

I also will need to update the toolchain files with an updated Arrow version and build recipe. 0.7.0 will probably be released next week roughly so you shouldn't have to be on an unreleased version for very long

wesm:
I'm out of gas for tonight. I have a few free hours tomorrow morning (Eastern Time) so I'll look at this first thing and and get the issues worked out

wesm:
@asuhan I'm finding when I build with gcc 4.9 that I get errors like

```
3: Test command: /home/wesm/code/mapd-core/build/Tests/ExecuteTest "--gtest_output=xml:../"
3: Test timeout computed to be: 9.99988e+06
3: F0907 11:00:11.645380 18703 QueryRunner.cpp:36] Check failed: boost::filesystem::exists(base_path) 
3: *** Check failure stack trace: ***
3:     @     0x7f45a03e2d5e  google::LogMessage::Fail()
3:     @     0x7f45a03e2caa  google::LogMessage::SendToLog()
3:     @     0x7f45a03e2690  google::LogMessage::Flush()
3:     @     0x7f45a03e5597  google::LogMessageFatal::~LogMessageFatal()
3:     @          0x120f8b6  get_session()
3:     @          0x11dadb3  main
3:     @     0x7f459c89ff45  __libc_start_main
3:     @          0x1174d9c  (unknown)
3:     @              (nil)  (unknown)
```

When I build with clang-3.8 instead, the tests work, but gdb doesn't work very well -- I'm able to get break points to work but inspecting variables seems not to work. I don't have any issues with gdb elsewhere, so it seems this is something MapD-specific. I can debug with print statements but gdb would really be better

wesm:
I created a `tmp` symlink to the `data` directory in `Tests` and seem to be in business. I must have followed the directions wrong someplace but let me know for the future

wesm:
Looking at this again later this evening

asuhan:
OK, I'll have a look myself tomorrow if you don't get to the bottom of it. FYI, I've seen a similar type of crash during my initial development of the test and it was due to buffer ownership problems. Holding the `PoolBuffer`s fixed it.

wesm:
Hm, interesting. Perhaps a pointer is being leaked somewhere. I will let you know if I can't sort it out

As an aside, it would also be better to use `std::shared_ptr<RecordBatch>`. That type is only copyable by accident. Just opened https://issues.apache.org/jira/browse/ARROW-1486

wesm:
OK, tests are passing again. I summarized the work in the commit message. 

Some follow up JIRAs

- https://issues.apache.org/jira/browse/ARROW-1493
- https://issues.apache.org/jira/browse/ARROW-1494

When you use `BufferReader` with the IPC read functions, so do not need to retain a reference to the original buffer, because internally the buffers in the record batch have a reference to the parent. See https://github.com/apache/arrow/blob/master/cpp/src/arrow/io/memory.cc#L227. If you had to keep track of which `arrow::Buffer` owns the memory all the time it would be complete madness. If you ever run into any problems with chained memory references, please bug me as I've tried to be really careful about simple zero-copy and RAII-based memory management but no one is perfect =) Better to fix the underlying issues than work around them.

wesm:
I have not made use of any of the direct-to-CUDA serialization code in Arrow yet. I think it would be better to scrutinize that patch more carefully in isolation

asuhan:
I agree we should leave the CUDA parts for later. We should also remove the dependency on the Arrow GPU, `scripts/common-functions.sh` doesn't build with GPU support, which breaks the CUDA build.

The patch looks good otherwise, I'll pick it with the dependency on Arrow GPU removed for now. Thanks!

wesm:
Oops, sorry about that (didn't test the toolchain from scratch). Thanks!

PR #92, opened 2017-09-24
fixed NULL from loading decimal string like ".99"

fleapapa:
Decimal constants without any digit before dot causes std::stoll to throws exception 'std::invalid_argument', making loaded decimal values become NULL.


asuhan:
Landed as https://github.com/mapd/mapd-core/commit/7f72560680dd3de8c1973073f1a0cd1b5d19e494.

PR #93, opened 2017-09-24
ENH: Add RPC endpoint for loading serialized Arrow columnar payloads

### LINKED ISSUE #89: Implement RPC endpoint for loading serialized Arrow record batches
e.g. `load_table_binary_arrow`. This would yield superior data import performance for clients that are able to create Arrow-formatted data in memory (see discussion in https://github.com/mapd/pymapd/pull/50)

I note that the codebase is accumulating a lot of usages of the `ENABLE_ARROW_CONVERTER` compiler definition. Would there be objections to make this a non-optional dependency? It is simple to build and does not introduce any additional transitive dependencies. We are quite committed to ensuring that the Arrow C++ libraries are a reliable thirdparty component for production applications, so if having this as a dependency in production caused you any problems, we would make fixing them a priority.

wesm:
There's some code duplication with the `load_table_binary_columnar`, but I don't have enough context about other parts of the system to know the best way to refactor. 

Some TODO

- [x] e2e testing with pymapd. Sanitize / cast types on client side for now
- [x] Throw exceptions instead of aborting when an Arrow column type is incorrect
- [x] check performance vs. load_table_binary_columnar from pymapd

This also makes Arrow a required build dependency

Close #89

### DIFF
--- a/Import/Importer.cpp
+++ b/Import/Importer.cpp
@@ -556,6 +556,12 @@ namespace detail {
 
 using namespace arrow;
 
+#define ARROW_THROW_IF(cond, message)           \
+  if ((cond)) {                                 \
+    LOG(ERROR) << message;                      \
+    throw std::runtime_error(message);          \
+  }
+
 template <typename ArrayType, typename T>
 inline void appendArrowPrimitive(const Array& values, const T null_sentinel, std::vector<T>* buffer) {
   const auto& typed_values = static_cast<const ArrayType&>(values);
@@ -571,7 +577,7 @@ inline void appendArrowPrimitive(const Array& values, const T null_sentinel, std
 }
 
 static void appendArrowBoolean(const ColumnDescriptor* cd, const Array& values, std::vector<int8_t>* buffer) {
-  CHECK(values.type_id() == Type::BOOL);
+  ARROW_THROW_IF(values.type_id() != Type::BOOL, "Expected boolean col");
   const int8_t null_sentinel = inline_fixed_encoding_null_val(cd->columnType);
   const auto& typed_values = static_cast<const BooleanArray&>(values);
   buffer->reserve(typed_values.length());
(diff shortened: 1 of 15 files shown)


asuhan:
Thanks, looks reasonable at the first glance. I'll have a closer look tomorrow.

wesm:
This is about ~10x faster end-to-end in Python based on initial testing: https://github.com/mapd/pymapd/pull/53#issuecomment-331762835. 

Beyond this additional performance savings would come from a native C++ Thrift client for MapD (like https://github.com/cloudera/hs2client). This is honestly not that big of a project, and might be worth it long-term.

ThriftHandler/MapDHandler.cpp
```diff
@@ -1074,6 +1073,116 @@ void MapDHandler::load_table_binary_columnar(const TSessionId& session,
...
+    LOG(ERROR) << ex.error_msg;
+    throw ex;
+  }
+  check_table_load_privileges(session, table_name);
+  std::unique_ptr<Importer_NS::Loader> loader;
+  if (leaf_aggregator_.leafCount() > 0) {
+    loader.reset(new DistributedLoader(session_info, td, &leaf_aggregator_));
+  } else {
+    loader.reset(new Importer_NS::Loader(cat, td));
+  }
```
asuhan: We should probably extract the first ~20 lines of this method into a separate one since they do setup work and access checks which don't have anything to do with the format of the input data. I don't mind doing it myself after we land this, anyway.

Import/Importer.cpp
```diff
@@ -549,6 +552,275 @@ void TypedImportBuffer::pop_value() {
...
+
+  for (int64_t i = 0; i < typed_values.length(); i++) {
+    if (typed_values.IsNull(i)) {
+      buffer->push_back(null_sentinel);
+    } else {
+      switch (unit) {
+        case TimeUnit::SECOND:
+          buffer->push_back(static_cast<time_t>(raw_values[i]));
+          break;
+        case TimeUnit::MILLI:
```
asuhan: These conversions make me slightly nervous, I had to count the zeros. Using `std::chrono::time_point` feels very heavy-handed, maybe write 1B like 1000 * 1000 * 1000 or something? At least until we require a C++14 compiler, then we can use digit separators.

wesm:
Thanks @asuhan. I can make these fixes tomorrow while also looking at the failure in https://github.com/mapd/pymapd/pull/53#issuecomment-332010225

asuhan:
Ok, I wasn't aware about that failure. Yes, we need to fix it before landing the patch.

wesm:
Yes, let's definitely have this resolved tomorrow so no holdups for @tmostak demo on Wednesday

Import/Importer.cpp
```diff
@@ -549,6 +552,275 @@ void TypedImportBuffer::pop_value() {
...
+  for (int64_t i = 0; i < typed_values.length(); i++) {
+    if (typed_values.IsNull(i)) {
+      buffer->push_back(null_sentinel);
+    } else {
+      buffer->push_back(raw_values[i]);
+    }
+  }
+}
+
+static void appendArrowBoolean(const ColumnDescriptor* cd, const Array& values, std::vector<int8_t>* buffer) {
```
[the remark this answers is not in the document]
wesm: done

Import/Importer.cpp
```diff
@@ -549,6 +552,275 @@ void TypedImportBuffer::pop_value() {
...
+  for (int64_t i = 0; i < typed_values.length(); i++) {
+    if (typed_values.IsNull(i)) {
+      buffer->push_back(null_sentinel);
+    } else {
+      buffer->push_back(static_cast<int8_t>(typed_values.Value(i)));
+    }
+  }
+}
+
+static void appendArrowInt16(const ColumnDescriptor* cd, const Array& values, std::vector<int16_t>* buffer) {
```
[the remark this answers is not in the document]
wesm: done

Import/Importer.cpp
wesm: done

ThriftHandler/MapDHandler.cpp
wesm: done

wesm:
All set

PR #102, opened 2017-10-04
 Fix a heap corruption

fleapapa:
Fix heap corruption from unexpected file read after running out of cpu buffers and Buffer::reserve() has failed with SlabTooBig exception


dwayneberry:
closed with 7d17a5334b3e8db3

PR #107, opened 2017-10-11
enable COPY from .gz/bz2/zip/tar/tgz files through libarchive

fleapapa:


fleapapa:
NOTE: Boost:iostream does not work for tar/tgz/zip types!!

fleapapa:
The "universal decompressor" of libarchive does work!!

PR #110, opened 2017-10-14
enable COPY from .gz/bz2/zip/tar/tgz files thru libarchive static binary

fleapapa:
squashed 4 commits


fleapapa:
squashed 4 commits

PR #112, opened 2017-10-15
Implemented optional error percentage argument for APPROX_COUNT_DISTINCT

### LINKED ISSUE #29: Make `APPROX_COUNT_DISTINCT` error configurable
Add a second, optional parameter to `APPROX_COUNT_DISTINCT` to control the error rate of the underlying HyperLogLog structure.

Smyatkin-Maxim:
It closes #29 

Usage: `APPROX_COUNT_DISTINCT(expr, integer_literal)`,
where `integer_literal` is the desired error rate in percents. Should be >= 1.

Internally it calculates a bitmap precision to provide the desired relative
error and uses it instead of `g_hll_precision_bits`.

### DIFF
--- a/QueryEngine/CalciteAdapter.cpp
+++ b/QueryEngine/CalciteAdapter.cpp
@@ -31,13 +31,18 @@
 
 namespace {
 
-ssize_t get_agg_operand_idx(const rapidjson::Value& expr) {
+std::vector<size_t> get_agg_operand_idxs(const rapidjson::Value& expr) {
   CHECK(expr.IsObject());
   CHECK(expr.HasMember("agg"));
   const auto& agg_operands = expr["operands"];
   CHECK(agg_operands.IsArray());
-  CHECK(agg_operands.Size() <= 1);
-  return agg_operands.Empty() ? -1 : agg_operands[0].GetInt();
+  auto n_operands = agg_operands.Size();
+  CHECK(n_operands <= 2);
+  std::vector<size_t> operands;
+  for (size_t i = 0; i < n_operands; ++i) {
+    operands.push_back(agg_operands[i].GetInt());
+  }
+  return operands;
 }
 
 std::tuple<const rapidjson::Value*, SQLTypeInfo, SQLTypeInfo> parse_literal(const rapidjson::Value& expr) {
@@ -452,14 +457,20 @@ class CalciteAdapter {
     CHECK(expr_type.IsObject());
     const auto agg_kind = to_agg_kind(expr["agg"].GetString());
     const bool is_distinct = expr["distinct"].GetBool();
-    const auto operand = get_agg_operand_idx(expr);
-    const bool takes_arg{operand >= 0};
-    if (takes_arg) {
-      CHECK_LT(operand, static_cast<ssize_t>(scan_targets.size()));
+    const auto operands = get_agg_operand_idxs(expr);
+    std::shared_ptr<Analyzer::Expr> arg_expr{nullptr};
+    std::shared_ptr<Analyzer::Constant> err_rate_expr{nullptr};
+    if (operands.size() >= 1) {
+      CHECK_LT(operands[0], scan_targets.size());
+      arg_expr = scan_targets[operands[0]]->get_own_expr();
+      if (operands.size() == 2) {
+        CHECK_LT(operands[1], scan_targets.size());
+        err_rate_expr = std::dynamic_pointer_cast<Analyzer::Constant>(scan_targets[operands[1]]->get_own_expr());
+        CHECK(err_rate_expr != nullptr);
+      }
     }
-    const auto arg_expr = takes_arg ? scan_targets[operand]->get_own_expr() : nullptr;
     const auto agg_ti = get_agg_type(agg_kind, arg_expr.get());
-    return makeExpr<Analyzer::AggExpr>(agg_ti, agg_kind, arg_expr, is_distinct);
+    return makeExpr<Analyzer::AggExpr>(agg_ti, agg_kind, arg_expr, is_distinct, err_rate_expr);
   }
 
   std::shared_ptr<Analyzer::Expr> translateTypedLiteral(const rapidjson::Value& expr) {
--- a/QueryEngine/CalciteAdapter.cpp
+++ b/QueryEngine/CalciteAdapter.cpp
@@ -31,18 +31,13 @@
 
 namespace {
 
-std::vector<size_t> get_agg_operand_idxs(const rapidjson::Value& expr) {
+ssize_t get_agg_operand_idx(const rapidjson::Value& expr) {
   CHECK(expr.IsObject());
   CHECK(expr.HasMember("agg"));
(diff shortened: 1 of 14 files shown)


QueryEngine/CalciteAdapter.cpp
```diff
@@ -452,14 +457,20 @@ class CalciteAdapter {
...
+      if (operands.size() == 2) {
+        CHECK_LT(operands[1], scan_targets.size());
+        err_rate_expr = std::dynamic_pointer_cast<Analyzer::Constant>(scan_targets[operands[1]]->get_own_expr());
+        CHECK(err_rate_expr != nullptr);
+      }
     }
-    const auto arg_expr = takes_arg ? scan_targets[operand]->get_own_expr() : nullptr;
     const auto agg_ti = get_agg_type(agg_kind, arg_expr.get());
-    return makeExpr<Analyzer::AggExpr>(agg_ti, agg_kind, arg_expr, is_distinct);
+    return makeExpr<Analyzer::AggExpr>(agg_ti, agg_kind, arg_expr, is_distinct, err_rate_expr);
```
asuhan: Feel free to pass a `nullptr` here and discard the rest of the changes to the file. This is an old adapter we've used for migrating to Calcite before having the full relational algebra model; we should delete it.

QueryEngine/HyperLogLog.h
```diff
@@ -74,6 +74,12 @@ inline void hll_unify(T1* lhs, T2* rhs, const size_t m) {
   }
 }
 
+inline int hll_sz_for_rate(const int err_percent) {
+  double err_rate{static_cast<double>(err_percent) / 100.0};
+  double k = ceil(2 * log2(1.04 / err_rate));
+  return k > 16 ? 16 : k < 1 ? 1 : k;
```
asuhan: Maybe something like `min(16, max(k, 1))` instead? The nested ternary is hard to read.

QueryEngine/HyperLogLog.h
```diff
@@ -74,6 +74,12 @@ inline void hll_unify(T1* lhs, T2* rhs, const size_t m) {
   }
 }
 
+inline int hll_sz_for_rate(const int err_percent) {
```
asuhan: Let's be more generous and use vowels: `hll_size_for_rate` or something similar.

QueryEngine/RelAlgTranslator.cpp
```diff
@@ -178,14 +178,22 @@ std::shared_ptr<Analyzer::Expr> RelAlgTranslator::translateAggregateRex(
     const std::vector<std::shared_ptr<Analyzer::Expr>>& scalar_sources) {
   const auto agg_kind = rex->getKind();
   const bool is_distinct = rex->isDistinct();
-  const auto operand = rex->getOperand();
-  const bool takes_arg{operand >= 0};
+  const bool takes_arg{rex->size() > 0};
+  std::shared_ptr<Analyzer::Expr> arg_expr = nullptr;
```
asuhan: You don't need the initialization to `nullptr`, the default constructor of `std::shared_ptr` does the right thing.

QueryEngine/RelAlgTranslator.cpp
```diff
@@ -178,14 +178,22 @@ std::shared_ptr<Analyzer::Expr> RelAlgTranslator::translateAggregateRex(
...
+  std::shared_ptr<Analyzer::Constant> err_rate = nullptr;
   if (takes_arg) {
+    const auto operand = rex->getOperand(0);
     CHECK_LT(operand, static_cast<ssize_t>(scalar_sources.size()));
+    arg_expr = scalar_sources[operand];
+    if (agg_kind == kAPPROX_COUNT_DISTINCT && rex->size() == 2) {
```
asuhan: Maybe check if `rex->size()` is 1 or 2.

QueryEngine/RelAlgOptimizer.cpp
```diff
@@ -760,11 +762,21 @@ std::vector<std::unique_ptr<const RexAgg>> renumber_rex_aggs(std::vector<std::un
                                                              const std::unordered_map<size_t, size_t>& new_numbering) {
   std::vector<std::unique_ptr<const RexAgg>> new_exprs;
   for (auto& expr : agg_exprs) {
-    auto old_idx = expr->getOperand();
-    auto idx_it = new_numbering.find(old_idx);
-    if (idx_it != new_numbering.end()) {
-      new_exprs.push_back(
-          boost::make_unique<RexAgg>(expr->getKind(), expr->isDistinct(), expr->getType(), idx_it->second));
+    bool need_renumber = false;
```
asuhan: Couldn't we just ignore the second argument and undo most of the changes here? It's always a literal anyway.

Smyatkin-Maxim: @asuhan I've seen you saying the same in the ticket, but wasn't sure why should I handle constant expressions differently. Perhaps, I don't completely understand how renumbering works.
Caught a cold a little bit yesterday, will look into it again as soon as I get better, probably day or two.

Smyatkin-Maxim: Even if literals indeed never get renumbered, can't use the old code because rex size now may be between 0 and 2 (0 operands for count(*), it was '-1' operand idx before).
If you don't like unnecessary std::vector allocations and redundant check in hashmap for literal idx, here how I can rewrite the function:

```c
std::vector<std::unique_ptr<const RexAgg>> renumber_rex_aggs(std::vector<std::unique_ptr<const RexAgg>>& agg_exprs,
                                                             const std::unordered_map<size_t, size_t>& new_numbering) {
  std::vector<std::unique_ptr<const RexAgg>> new_exprs;
  for (auto& expr : agg_exprs) {
    if (expr->size() >= 1) {
      auto old_idx = expr->getOperand(0);
      auto idx_it = new_numbering.find(old_idx);
      if (idx_it != new_numbering.end()) {
        std::vector<size_t> operands;
        operands.push_back(idx_it->second);
        if (expr->size() == 2) {
          operands.push_back(expr->getOperand(1));
        }
        new_exprs.push_back(boost::make_unique<RexAgg>(expr->getKind(), expr->isDistinct(), expr->getType(), operands));
        continue;
      }
    }
    new_exprs.push_back(std::move(expr));
  }
  return new_exprs;
}
```

I'd probably prefer the current version because:
1. it will work whatever happens in other parts of code (if literals get renumbered some other day, or perhaps we add another aggregate with two arguments)
2. this function is far from hot path, so one allocation and one lookup probably won't make too big a difference.

@asuhan what do you think?

asuhan: I'm not convinced that calling renumbering on literals is a legitimate use of this code. I'd rather have it fail quickly if the caller attempts something dumb. I have no other concerns, of course the allocation and lookup you do here are completely fine. Anyway, the patch looks good overall and I'm not going to hold it because this renumbering change, at least it looks correct.

asuhan:
Good work, thanks!

PR #116, opened 2017-10-16
Fixed JDBC ignoring prepared statement parameter at the end of a query

### LINKED ISSUE #45: JDBC PreparedStatement won't work with parameter as last element
Using release 3.1.1 and the associated jdbc driver lib.

```
CREATE TABLE t (x INT NOT NULL);
```

In Java code, assuming properly set up connection:
```
PreparedStatement ps = conn.prepareStatement("SELECT x FROM t WHERE x = ?");
ps.setInt(1, 1)
```
fails with:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 0
	at com.mapd.jdbc.MapDPreparedStatement.setInt(MapDPreparedStatement.java:175)
```

This is due to the `MapDPreparedStatement` at https://github.com/mapd/mapd-core/blob/master/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDPreparedStatement.java#L84 initially splitting the SQL String using `.split`. The javadoc for that method states:
> Trailing empty strings are therefore not included in the resulting array. 

Therefore the subsequent line of code calculates the number of expected parameters as zero.

A simple workaround is to include an additional space at the end of the statement i.e. `"SELECT x FROM t WHERE x = ? "` but that shouldn't really be necessary.

Smyatkin-Maxim:
Closes #45 

To reproduce: "SELECT * FROM t WHERE id = ?"
Negative "limit" argument in String.split() prevents it from ignoring empty splits

### DIFF
--- a/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDPreparedStatement.java
+++ b/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDPreparedStatement.java
@@ -82,7 +82,7 @@ class MapDPreparedStatement implements PreparedStatement {
     this.stmt = new MapDStatement(session, client);
     MAPDLOGGER.debug("Prepared statement is " + currentSQL);
     //TODO in real life this needs to check if the ? isinside quotes before we assume it a parameter
-    brokenSQL = currentSQL.split("\\?");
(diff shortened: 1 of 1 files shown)


dwayneberry:
Thanks, I have merged the change

PR #122, opened 2017-10-20
Fix reverse_logical_distribution corner cases

asuhan:
Handle the case where nothing is left after extracting the common expression.


dwayneberry:
thanks

PR #129, opened 2017-11-21
define Calcite in README fix #128

### LINKED ISSUE #128: define `Calcite` in README
the [README](https://github.com/mapd/mapd-core/blob/master/README.md) refers to `Calcite`, but doesn't explicitly define what `Calcite` is.

It would be nice to add a definition of `Calcite` in the context of **mapd-core** to the README.

micahstubbs:
A small PR to define [Calcite](http://calcite.apache.org/)  (it's a SQL Parser!) and make the dev docs a little more accessible to new **mapd-core** users

fix #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The `sanity_tests` target runs the most common tests. If using Makefiles to buil
 
 ## AddressSanitizer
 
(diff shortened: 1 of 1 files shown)


dwayneberry:
@micahstubbs I am just removing the references to calcite from the readme so I wouldn't go any further with this

micahstubbs:
looks like this one is obsolete, will close this PR then.

PR #132, opened 2017-11-25
SQL auto-complete endpoint

asuhan:


java/calcite/src/main/java/com/mapd/parser/server/CalciteServerHandler.java
```diff
@@ -164,4 +171,72 @@ public void updateMetadata(String catalog, String table) throws TException {
...
+      String msg = "Could not get Parse Item from pool: " + ex.getMessage();
+      MAPDLOGGER.error(msg);
+      throw new TException(msg);
+    }
+    MapDUser mapDUser = new MapDUser(user, session, catalog, mapdPort);
+    MAPDLOGGER.debug("getCompletionHints was called User: " + user + " Catalog: " + catalog + " sql: " + sql);
+    parser.setUser(mapDUser);
+
+    // remove last character if it is a ;
+    if (sql.charAt(sql.length() - 1) == ';') {
```
dwayneberry: we need a .trim sql prior to this, not that it makes sense to be calling for completion with a full sql stmt anyway.

It probably makes sense in this context to remove this code to remove `;` entirely

asuhan: I'm pretty sure neither is required, I remember testing with whitespace around the query. i'll remove it.

Issue #73, opened 2017-09-18
Implement ROUND function


asuhan:
We have `ROUND_TO_DIGIT` which is meant to serve the same purpose, but it isn't precise (converts the first argument to double) and its return type is always `DOUBLE`.

psockali:
Hi, is it OK if I try to close this issue?

billmaimone:
Sure, go ahead and assign yourself.

psockali:
I have  been going through the comments in #72 
According to @asuhan the `ROUND` function shall be implemented in a similar fashion to `FLOOR` in that it does retain the `typeinfo` of the first operand. Will this be a reasonable path to a solution?

billmaimone:
Yes, noting also commonly-adopted semantics of the optional second argument.

psockali:
managed to get the round function working.
Was wondering about following use-case:

```
create table table_a (
  val1 DECIMAL(10,5),
  val2 INT
)

SELECT ROUND(val1, val2) FROM table_a;

```

Is this a valid use case?
Such a query is currently possible.

billmaimone:
Useful bonus points to accept an integer column also for the second column, but uncommon in practical usage.  For the implementation did you consider using an existing library function for the implementation?

psockali:
Yes, I looked at the cmath functions, but none of them allowed me to round a value according to the requirements. I will keep an open eye on this as I am pretty sure this has most likely already been done.

psockali:
been thinking about the  column as 2nd parameter use case.
If this is practically never used, I would prefer to disable it.
The rational is: how should it behave if one of the entries in the column is NULL ?
Before going into length to address this allowing only constants / literals seems rather straight forward.

psockali:
Tried it out after looking a bit at #2 . It actually already behaves correctly in that for NULL 2nd operands a NULL is returned. And exactly this behaviour is required, because of both integer columns as well as sub-queries.

psilvest:
This issue has been inactive for a while, and the project has changed substantially. If this still occurs on the latest release, please add reproduction steps and environment details.

PR #139, opened 2017-12-17
ROUND function implementation

psockali:
Initial implementation proposal for issue #73 
Please start reviewing to check if this is an acceptable approach, I shall add the test cases tomorrow (only checked manually via madql).


psockali:
@billmaimone initial test case added and it passes locally

VrajPandya:
Hello @psockali Thanks for your contribution. Two quick small things:

Please add cmath header when you use pow function old compilers can't pick it up.

If you do ROUND(-1.7, -1) it gives -0 as a result we would like it to be 0.

Thanks very much.

psockali:
Hey @VrajPandya 

I have addressed the issues mentioned by you.
Please review again.

Cheers

PR #142, opened 2017-12-21
Switch to token-based completion hints

asuhan:
Calcite doesn't honor our permissions and the context-sensitive identifier
completion is brittle. Only use it for keywords, filtered by a narrow
whitelist.

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -40,6 +40,7 @@ add_executable(StorageTest StorageTest.cpp PopulateTableRandom.cpp ScanTable.cpp
 add_executable(StoragePerfTest StoragePerfTest.cpp PopulateTableRandom.cpp ScanTable.cpp)
 add_executable(ImportTest ImportTest.cpp QueryRunner.cpp)
 add_executable(TopKTest TopKTest.cpp QueryRunner.cpp)
+add_executable(TokenCompletionHintsTest TokenCompletionHintsTest.cpp)
 
 target_link_libraries(PlanTest gtest ${MAPD_LIBRARIES} ${Boost_LIBRARIES} ${Glog_LIBRARIES} ${CMAKE_DL_LIBS} ${CUDA_LIBRARIES} ${LLVM_LINKER_FLAGS} ${CURSES_LIBRARIES})
 target_link_libraries(ProfileTest gtest ${MAPD_LIBRARIES} ${Boost_LIBRARIES} ${Glog_LIBRARIES} ${CMAKE_DL_LIBS} ${CUDA_LIBRARIES} ${PROF_LIBRARIES} ${LLVM_LINKER_FLAGS} ${CURSES_LIBRARIES})
@@ -55,6 +56,7 @@ target_link_libraries(RunQueryLoop ${EXECUTE_TEST_LIBS})
 target_link_libraries(StringDictionaryTest StringDictionary gtest ${Boost_LIBRARIES})
 target_link_libraries(ImportTest gtest ${EXECUTE_TEST_LIBS})
 target_link_libraries(TopKTest ${EXECUTE_TEST_LIBS})
+target_link_libraries(TokenCompletionHintsTest token_completion_hints gtest mapd_thrift ${Boost_LIBRARIES})
 
 set(TEST_ARGS "--gtest_output=xml:../")
 add_test(PlanTest PlanTest ${TEST_ARGS})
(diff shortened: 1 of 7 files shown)


micahstubbs:
what would it take to add a `callback` argument to `get_completion_hints` at this line? https://github.com/mapd/mapd-core/blob/alex/completion_improvements/ThriftHandler/MapDHandler.h#L160

I'd like it to accept a callback like `render_vega` does here https://github.com/mapd/mapd-core/blob/alex/completion_improvements/ThriftHandler/MapDHandler.h#L183

micahstubbs:
ok, will share my efforts to test the `get_completion_hints` endpoint using the **mapd-connector** javascript client

first I look at `mapd-core/ThriftHandler/MapDHandler.cpp` to figure out what arguments `get_completion_hints` accepts:
screen shot 2018-01-03 at 3 02 09 pm

then I write a function in `mapd-connector/src/mapd-con-es6.js` to synchronously call `get_completion_hints` with the arguments that it expects:
screen shot 2018-01-03 at 3 01 51 pm

with that function in hand, I build **mapd-connector**, commit the changes, push them, pull them down into the `mapd-immerse/libs/connector` directory, run Immerse, and open the browser console to test out the new **mapd-connector** `getCompletionHints` endpoint.  here is what I see when I do this, for a variety of `queryString, cursorPosition` input:
screen shot 2018-01-03 at 3 02 31 pm

strangely, all requests in this test return an empty array.  why is this. can you see what I am missing?

tmostak:
Can you try decreasing the the index by one? Not sure of the API here but perhaps its supposed to be the 0-based index of the character the cursor is after.  So something like getCompletionHints('SEL', 2).

micahstubbs:
@tmostak good thinking. 

when I use a 0-based cursorPosition index, we still see an empty array response 🤔 

screen shot 2018-01-03 at 4 01 48 pm

 must be another issue somewhere with the arguments, my client, or the service

micahstubbs:
I get a good completionHint response with these parameters @andrew suggests in both the *mapd-connector* browser javascript client

`['select co', -1]`

screen shot 2018-01-03 at 4 30 49 pm

micahstubbs:
to verify this, I passed the same `['select co', -1]` **queryString**, **cursorPosition** parameters to the custom python thrift client that @andrewseidl made and sent to me.

screen shot 2018-01-03 at 4 26 28 pm

in case you are curious, we are passing `"random garbage ignored"` as the session ID, which is either ignored or has some other, non-blocking problem right now.

micahstubbs:
learning more:  a **cursorPosition** value of `-1` means  `end of current statement`

micahstubbs:
@asuhan found some edge cases that would probably be worth a look:

using @andrewseidl's custom python thrift client for MapD, I try this call

```
./MapD-remote -h bewdy.mapd.com:39091 get_completion_hints "session_id"  "select fr" -1
```

I'd expect to see a response like:
```
[TCompletionHint(type=7, hints=['FROM'], replaced='fr')]
```

micahstubbs:
it also seems like we can improve how we parse special characters like `*`

when I try this query, I see this error in the `mapd-server` logs:

query:

error:
```
2018-01-03 18:40:26 ERROR CalciteException:<init>:59 - org.apache.calcite.sql.validate.SqlValidatorException: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:61 - org.apache.calcite.runtime.CalciteContextException: At line 1, column 8: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:59 - org.apache.calcite.sql.validate.SqlValidatorException: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:61 - org.apache.calcite.runtime.CalciteContextException: At line 1, column 8: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:59 - org.apache.calcite.sql.validate.SqlValidatorException: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:61 - org.apache.calcite.runtime.CalciteContextException: At line 1, column 8: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:59 - org.apache.calcite.sql.validate.SqlValidatorException: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteException:<init>:61 - org.apache.calcite.runtime.CalciteContextException: At line 1, column 8: Unknown identifier '*'
2018-01-03 18:40:26 ERROR CalciteServerHandler:getCompletionHints:196 - Could not retrieve completion hints: null
```

micahstubbs:
a comment from @asuhan via slack

> I think Calcite doesn't want to auto-complete `SELECT` and `FROM` keywords, but we can easily add them...so it has nothing to do with `*`

micahstubbs:
looking at this more, it seems like `get_completion_hints` doesn't want to provide hints for substrings at the end of a queryString.  look at this set of calls, for example:

```
(hints_client_env_py36) ➜  mapd ./MapD-remote -h bewdy.mapd.com:39091 get_completion_hints "oO6wYy9qR2vZUaK3fO3X58hYa0Imc9Il"  "select bo fr sdf" 12
[TCompletionHint(type=7, hints=['FROM'], replaced='fr')]
(hints_client_env_py36) ➜  mapd ./MapD-remote -h bewdy.mapd.com:39091 get_completion_hints "session_id"  "select co fr sdf" 12
[TCompletionHint(type=7, hints=['FROM'], replaced='fr')]
(hints_client_env_py36) ➜  mapd ./MapD-remote -h bewdy.mapd.com:39091 get_completion_hints "session_id"  "select fr" -1
[]
(hints_client_env_py36) ➜  mapd ./MapD-remote -h bewdy.mapd.com:39091 get_completion_hints "session_id"  "select * fr" -1
[]
(hints_client_env_py36) ➜  mapd
```

in screenshot form:
screen shot 2018-01-04 at 11 20 48 am

micahstubbs:
ok, I tested the fixes in the last two commits and verified that they work as advertised.  this PR is ready to merge as far as I am concerned.  +1

PR #145, opened 2017-12-28
issues/2: adding hybrid solution for hoisting constants

psockali:
This is the initial hybrid solution.
It is far from perfect, but it hoisted constants correctly (in most cases, except for CASE) into the query functions .entry block.

### DIFF
--- a/QueryEngine/NativeCodegen.cpp
+++ b/QueryEngine/NativeCodegen.cpp
@@ -1048,8 +1048,7 @@ Executor::CompilationResult Executor::compileWorkUnit(const std::vector<InputTab
   bind_pos_placeholders("pos_step", false, query_func, cgen_state_->module_);
 
   cgen_state_->query_func_ = query_func;
-  cgen_state_->query_func_entry_ir_builder_.SetInsertPoint(&query_func->front(),
-                                                           query_func->front().getInstList().begin());
+  cgen_state_->query_func_entry_ir_builder_.SetInsertPoint(&query_func->getEntryBlock().front());
 
   std::vector<llvm::Value*> col_heads;
   std::tie(cgen_state_->row_func_, col_heads) = create_row_function(ra_exe_unit.input_col_descs.size(),


QueryEngine/ConstantIR.cpp
```diff
@@ -96,38 +96,168 @@ llvm::ConstantInt* Executor::codegenIntConst(const Analyzer::Constant* constant)
...
+}
+
 std::vector<llvm::Value*> Executor::codegenHoistedConstants(const std::vector<const Analyzer::Constant*>& constants,
                                                             const EncodingType enc_type,
                                                             const int dict_id) {
   CHECK(!constants.empty());
   const auto& type_info = constants.front()->get_type_info();
-  auto lit_buff_lv = get_arg_by_name(cgen_state_->row_func_, "literals");
-  int16_t lit_off{-1};
+  int64_t lit_off{-1};
```
shtilman: Could you describe why you decided to revamp this function? I see different versions.. What is the workflow?

psockali: - codegenHoistedConstantsInPlace
  Creates the literal load from the literals buffer in place, effectively the original codegenHoistedConstants enhanced to also set the names for IR nodes accordingly
- codegenHoistedConstants 
  This is the original entry point, it has been changed to first check if the Constant was already hoisted (by checking presence in the function argument list). If not, it calls codegenHoistedConstantsInPlace to generate the load.
- codegenHoistedConstantsInEntryBlock (just renamed it)
  New entry point for the generation of the loads in the .entry block. It only creates the load, if the Constand had not been hoisted before.

psockali:
Hey @shtilman , 

Sorry for the poor documentation on the change... will try to avoid this in the future. 

The basic idea was as follows:

1. Scan the RA Tree for Constants via the visitor 
2. Generate loads for found and valid Constants into the .entry block
    This is being done directly while scanning the RA Tree, because sometimes the parent RA node is required for correct type information for code generation. Doing this only on the Constants caused quite a few Tests to fail (especially where varchars were involved).
3. Use the .entry block's IR as an indirect form of mapping
    The gen context already maintains a Constant -> offset mapping
4. and finally when generating row_func it looks up the the right argument by name and, if found, uses it, otherwise generates a new load from the literals buffer.

This only works because duplicate Constants are eliminated during code generation, so the offset into the literals buffer is identical for Constants equal by value, and thus the names can be used for looking them up.

The rational for using the .entry block directly as mapping was my concern with the IN operator.
It has a hard limit of 5 million entries, so was trying to avoid maintaining too many mappings for the same thing (as the de-duplication already does this in some form: Constant -> offset ).

The hoist_literals flag is still being considered. If it is not set, then no constants will be hoisted.
This is achieved by passing the relevant methods a NULL instead of an instance of ConstantHoistingVisitor.

Please advice how to proceed.

shtilman:
Ok, let me dig a bit deeper into your approach.

psockali:
I had a revisit as well and I think your are right to introduce a mapping between Constans (or their offset) and the function arguments / .basic_block literals load instructions.

It turns out that I am currently doing a linear search (by name) for the function arguments and in the .entry block for hoisted literal loads. I was hoping to do a binary search, but sadly llvm::SymbolTableList is a linked list, which sort of diminishes the benefits of binary search.

Please bear with me until I make the amendments.

asuhan:
Why would you even care about a binary search? We're talking maybe 100 constants at most.

psockali:
Maybe I am missing something, but looking at the `RelAlgTranslator::translateInOper` I had the impression that it can create quite a few Constants instances. There is a special path for integers and dict encoded strings, but for decimals and floats it would take the hit and create as many Constants as there are results in the result set. Let me see if I can create a test case for this scenario.

asuhan:
Sure, but there is a lot of linear-time work to be done anyway during code generation, not sure what the impact would be. To be honest, I'm not entirely convinced we'd survive generating a kernel with enough constants for binary search to make a difference anyway. Also, in terms of use cases, the right hand side of an `IN` expression should be something which makes sense to enumerate (user names or ids, for example), which isn't really the case for floating point and decimals. So I'm not particularly bothered by the lack of binary search and would rather take simpler code.

psockali:
Yes, makes sense. In that case I am going to leave it as it is for now, so as not to create too much deltas for a review.

psockali:
Hey @shtilman,

this whole 'Spagat' only made sense with the assumption that there might be up to 5 million Constants (by using the `IN` operator). Now that assumption has been rendered void. It all feels like doing an initial pass to capture things and a final pass to generate the target IR seems most reasonable.
How do you feel about trashing this PR and instead refocusing on two/multi pass based IR generation, with constants hoisting as the initial benefactor of the multi-pass paradigm?

asuhan:
In which situation we care about could the `IN` operator create over 5 million constants? For integer and strings we generate a bitmap instead. Can we throw an exception if the `IN` operator generates 5+ million constants and still work under the same assumption?

psockali:
@asuhan, I am not sure I am following you correctly.

The check is already there (https://github.com/mapd/mapd-core/blob/647a0efd7d931aa27105652796aa7416f19b1dc4/QueryEngine/RelAlgTranslator.cpp#L369), but I am not sure if g_enable_watchdog is true.

And I thought that decimal/floats are virtually not a use case.

Are you saying that a single pass with a retrospective mutation of the query_function's declaration (including a search and replace of the query_function's IR) outweighs the complexity it introduces?

I mean, if there are at most about 100 hoisted literals, I doubt it would make a big difference.

Either way, please advise if I should attempt the single pass approach...

shtilman:
@psockali it seems to me that we need to tie constant hoisting to the actual [staged] codegen, which would solve the problem of on-the-fly constants and also subquery constants, since hoisting would be staged too. All relevant constants would be injected into each query's compilation, after all subqueries have been compiled and executed. 

My proposal is to continue filling up `literals_` buffer but defer literal load generation. Or, perhaps, redirect generation of literal loads to the query func's entry block, i.e. whenever a constant is codegened, `query_func`'s  entry block would get the load from literal buffer (if it hasn't yet been loaded) and  `row_func` would get an appropriately named arg value. You know the offset and the type of each literal so you can then connect the dots in the IR before compiling it.

If number of constants gets too big you could throw an exception or you could stop hoisting.

asuhan:
@psockali `g_enable_watchdog` is true by default, we only disable it for unit tests and various experiments. I'd rather reject the 5M+ case altogether (even with watchdog disabled) than make heroic efforts for a kernel which might even fail to compile or do so very slowly -- 5M+ instructions is *a lot*.

psockali:
But @asuhan, that is what I am saying. I misread the code. It actually does not create 5M load instructions, that is an upper limit for the elements in the bit-set. Either way, I see if I can get the modification of the function declaration / signature working.

asuhan:
@psockali In cases where a bitmap cannot be used (floats for example) it would. There would be 5M+ `Analyzer::Constant` expressions generating the same number of constants to be loaded from the literal buffer when the `IN` expression is generated.

shtilman:
@psockali modifying row_func signature may get messy. You may want to consider switching to global variables with internal linkage. For each new offset you could create a `literal_<offset>` global visible to the entire module (that is, if it hasn't been created yet - call `getNamedGlobal(name)` to check). If the module doesn't have it yet  - 1. create it  2. then define it with a load from `literals_` in `query_func()` entry block and  3. use it. If module already has it - skip steps 1. and 2.
Look at the resulting IR. Global optimizer should be able to handle these globals after row_func is inlined, they should be all localized.
If that works well - no need to open the Pandora's box messing with function signatures.

psockali:
@asuhan: Would it be OK if I disable the float/decimal use-case for the `IN` operator?
@shtilman: I was hoping to use the function copy helper to change the signature, but I shall try global variables first.

asuhan:
@psockali No it wouldn't, but it'd be ok to reject it past a certain threshold (5M is as good of a value as any).

psockali:
Just noticed a cast from `size_t` to `int16_t` when loading values from the literals buffer... 

https://github.com/mapd/mapd-core/blob/245f9ffe9e036f88171555c71e63d10827efd9b9/QueryEngine/ConstantIR.cpp#L110

I guess it really does not hold a lot of elements.

psockali:
@shtilman, i am not sure how to correctly initialize the `GlobalVariable`, shall I create a correct pointer type and use store/load? This then feels like copying the buffer into a .data segment and then loading the values from that .data segment for each use. Not sure if the optimizer will be able to detect this correctly.

psockali:
The cast actually limits the buffer to 32K, with the CHECK enforcing this, so there is then already a reasonable upper limit.

shtilman:
@psockali yes, I think a simple store into a global var should work. You first load the value from the literal buffer at specific offset, then store into corresponding global which is named after that offset - this is in the entry block. Then you can use that `literal_<offset>` global whenever you need `literals_[offset]`.

Make sure you create each global var in the query module, with *internal* linkage. After `row_func` is inlined, the access to each global you had created will be limited to a single non-recursive function. Then the GlobalOptimizer pass (which follows the inliner in the pipeline) should take all these *internal* globals and localize them, i.e. convert them into local allocas.

psockali:
@shtilman, I did as you said, but it complains at link time.

> LLVM ERROR: Program used external function 'global_literal_0' which could not be resolved!

psockali:
I have also pushed the changes

shtilman:
Have you looked at the IR that you end up generating?
Could you run `explain` on that failing query and post the result here?

psockali:
Yes, I did. The IR looks OK. The global variable is still present after the optimisation causing the linker to fails, because I am not passing in sufficient linkage information.

The query I used was: 

> select cast(-1.1 as int) as x from t2;

The generated bitcode is attached: [bitcode.zip](https://github.com/mapd/mapd-core/files/1676498/bitcode.zip)

psockali:
@shtilman I have committed now a version that alters the row_funcs signature (or more precise, creates a new row_func_hoisted fuction with the correct signature). This passes all the tests and the optimized IR looks good as well, as far as I can tell.

shtilman:
All right, was hoping globals would simplify the hoisting, provided globalopt worked well. Not sure why globalopt failed to localize them, there are a few straightforward conditions each global should meet, perhaps globalvar declarations needed a bit of massaging.. Don't have time to investigate, we'll do signature change.

QueryEngine/NativeCodegen.cpp
```diff
@@ -1077,6 +1081,101 @@ Executor::CompilationResult Executor::compileWorkUnit(const std::vector<InputTab
     }
   }
 
+  std::vector<llvm::Value*> hoisted_literals;
+
+  if (co.hoist_literals_ && !cgen_state_->defined_literals_.empty()) {
+    // we have some literals...
+
+    // create a new row_function with the literals as argument
```
shtilman: Need a better comment, something like
```
// row_func_ is using literals whose defs have been hoisted up to the query_func_,
// extend row_func_ signature to include extra args to pass these literal values. 
```

QueryEngine/NativeCodegen.cpp
```diff
@@ -1077,6 +1081,101 @@ Executor::CompilationResult Executor::compileWorkUnit(const std::vector<InputTab
...
+
+    for (auto& element : cgen_state_->defined_literals_) {
+      for (auto value : element.second) {
+        row_process_arg_types.push_back(value->getType());
+      }
+    }
+
+    auto ft = llvm::FunctionType::get(get_int_type(32, cgen_state_->context_), row_process_arg_types, false);
+    auto row_func_hoisted = llvm::Function::Create(
+        ft, llvm::Function::ExternalLinkage, "row_func_hoisted", cgen_state_->row_func_->getParent());
```
shtilman: Maybe give it a clearer name, e.g. `"row_func_hoisted_literals"` ?

QueryEngine/ConstantIR.cpp
```diff
@@ -115,19 +117,87 @@ std::vector<llvm::Value*> Executor::codegenHoistedConstants(const std::vector<co
     }
   }
   CHECK_GE(lit_off, int16_t(0));
-  const auto lit_buf_start = cgen_state_->ir_builder_.CreateGEP(lit_buff_lv, ll_int(lit_off));
+
+  int64_t allocated_watermark = cgen_state_->literal_bytes_high_watermark(0);
+  std::string literal_name = "literal_" + std::to_string(lit_off);
+
+  if (allocated_watermark == initial_watermark) {
```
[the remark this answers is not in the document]
psockali: This was just some form of precaution.
If something was already present in the literals buffer, I expect it to be present in the defined_literals map. I change it to only check against the map.

QueryEngine/NativeCodegen.cpp
```diff
@@ -1047,6 +1047,10 @@ Executor::CompilationResult Executor::compileWorkUnit(const std::vector<InputTab
   bind_pos_placeholders("group_buff_idx", false, query_func, cgen_state_->module_);
   bind_pos_placeholders("pos_step", false, query_func, cgen_state_->module_);
 
+  cgen_state_->query_func_ = query_func;
+  cgen_state_->query_func_entry_ir_builder_.SetInsertPoint(&query_func->front(),
```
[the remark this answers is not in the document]
psockali: The create_row_func() method already adds some instructions, thus I am using this mean to ensure that literals resolution ends up as the first instructions.

QueryEngine/NativeCodegen.cpp
psockali: yes, will do

QueryEngine/NativeCodegen.cpp
psockali: yes, makes sense

QueryEngine/NativeCodegen.cpp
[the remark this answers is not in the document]
shtilman: Sure. What I meant was replacing `query_func->front()` with  `query_func->getEntryBlock()`.
You could point that irbuilder to the very beginning of the entry block like this:
```
cgen_state_->query_func_entry_ir_builder_.SetInsertPoint(&query_func->getEntryBlock().front());
```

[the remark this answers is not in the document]
psockali: changed it accordingly

PR #149, opened 2018-01-04
Decompress dictionary-encoded sides when encoding is different

asuhan:
Commit f2a2b5b41b7b347ed9e85be7509b2472632a9598 changed the logic to
account for dictionaries with the same encoding, but it didn't check
that the dictionaries are actually shared (same compression parameter).


VrajPandya:
The changes seems fine.

PR #181, opened 2018-01-21
Disable sharding for baseline outer joins as well

asuhan:
The fragment distribution logic must change a bit before outer joins can
be sharded. Disable for now for correctness' sake, which wasn't done for
the baseline join hash table layout (used for joins on multiple columns).


asuhan:
The failure of the newly added test was triggered when running tests with sharding (the `--with-sharding` flag), on a machine with more than one GPU.

PR #199, opened 2018-02-25
[WIP] Upgrade to Apache Arrow >= 0.8.0 API 

wesm:
I ran into some build problems in https://github.com/mapd/mapd-core/issues/198 and am not able to build the whole project yet. I will be happy to complete this once that is resolved

### DIFF
--- a/QueryEngine/ResultSetConversion.cpp
+++ b/QueryEngine/ResultSetConversion.cpp
@@ -336,7 +336,7 @@ std::shared_ptr<arrow::RecordBatch> ResultSet::getArrowBatch(const std::shared_p
 
   const auto entry_count = entryCount();
   if (!entry_count) {
-    return std::make_shared<arrow::RecordBatch>(schema, 0, result_columns);
+    return arrow::RecordBatch::Make(schema, 0, result_columns);
   }
   const auto col_count = colCount();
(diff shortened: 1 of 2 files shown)


andrewseidl:
Do you see any issues with adding an `#ifdef` to keep 0.7.x compatibility around temporarily (and maybe warn as deprecated)? Simply so that we don't have to force everyone to upgrade their dev envs immediately.

Last I looked it didn't seem like Arrow exposed its version in the headers, so we could instead either attempt to pick up the version through pkg-config, or (preferably) do a cmake `try_compile()` to decide which one to use (or maybe `CheckCXXSymbolExists`).

Should be quick, I can take care of it hopefully today.

wesm:
Applications using 0.8.0 are binary-incompatible with <= 0.7.1, so there's only so much benefit to maintaining the old APIs.

randyzwitch:
@andrewseidl Is this something that will make it in soon? Binary incompatibility is making development weird

wesm:
Upgrading should not be difficult. I don't have the bandwidth to steward through a patch right now; if someone can take it over, building the project from source and running the tests is the hardest part

randyzwitch:
Thanks @andrewseidl, let me know what I can do to help if needed

PR #221, opened 2018-06-05
[WIP] Resolves #206 - Added new data types support

### LINKED ISSUE #206: gdf/arrow: missing support for decimal, date, time, timestamp
We currently only support bool, int*, float*, and text when returning results using Arrow. Selecting decimal, date, time, or timestamp columns isn't currently supported and returns a somewhat confusing exception:

    <type> is not supported in temporary table.

Type convert and the data conversion/column builder is around https://github.com/mapd/mapd-core/blob/dd6c3c6482630f3794c6e4cbd9783e70fd4436f3/QueryEngine/ResultSetConversion.cpp#L151-L185

cc: https://github.com/mapd/pymapd/issues/75
cc: https://github.com/mapd/pymapd/issues/76

xmnlab:


xmnlab:
Is there a CI for mapd-core?
is there a documentation for contributors? or should we just try to recompile locally?

PR #276, opened 2018-10-24
jdbc: generate recognized syntax when using SetBoolean

### LINKED ISSUE #274: JDBC setBoolean implementation
Hi there! 
Why does the implementation of `setBoolean` in JDBC drive converts values to `t` or `f`? I get an exception `java.sql.SQLException: Query failed : Exception: Cannot compare between BOOLEAN and VARCHAR`. 

I'm using `mapdjdbc-1.0-SNAPSHOT-jar-with-dependencies.jar` and as I see, implementation is the same in master branch, https://github.com/omnisci/mapd-core/blob/master/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDPreparedStatement.java#L167 

Example code:
```scala
  val sql  = s"select 1 from example where hasException = ?"
  val stmt = ConnectionPool.get().borrow().prepareStatement(sql)
  stmt.setBoolean(1, false)
  val resultSet = stmt.executeQuery()
  println(resultSet.getInt(1))
```
```bash
22:22:24.092 [run-main-6] DEBUG com.mapd.jdbc.MapDStatement - sql is :'select 1 from example where hasException = 'f' LIMIT 100000'
22:22:24.092 [run-main-6] DEBUG com.mapd.jdbc.MapDStatement - afterFnSQL is :'select 1 from example where hasException = 'f' LIMIT 100000'
```

I obtain connection via ScalikeJDBC's `ConnectionPool`, but it's the same as `DriverManager.getConnection(...)`

```scala
  Class.forName("com.mapd.jdbc.MapDDriver")
  ConnectionPool.singleton("jdbc:mapd:192.168.1.9:9091:db_test", "mapd", "HyperInteractive")
```

Am I doing something wrong?

andrewseidl:
Previously the generated sql was of the form:

    WHERE column = "f"

which is not recognized by Calcite. Now it generates:

    WHERE column = false

Resolves #274

### DIFF
--- a/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDPreparedStatement.java
+++ b/java/mapdjdbc/src/main/java/com/mapd/jdbc/MapDPreparedStatement.java
@@ -166,8 +166,8 @@ class MapDPreparedStatement implements PreparedStatement {
   @Override
   public void setBoolean(int parameterIndex, boolean x) throws SQLException {
     MAPDLOGGER.debug("Entered");
-    parmRep[parameterIndex - 1] = x ? "t" : "f";
-    parmIsString[parameterIndex - 1] = true;
+    parmRep[parameterIndex - 1] = x ? "true" : "false";
+    parmIsString[parameterIndex - 1] = false;
     parmIsNull[parameterIndex - 1] = false;
     repCount++;
   }


andrewseidl:
Sample usage:

```
import java.sql.*;

public class brokenjdbc {
  static final String JDBC_DRIVER = "com.mapd.jdbc.MapDDriver";
  static final String DB_URL = "jdbc:mapd:localhost:9091:mapd";

  static final String USER = "mapd";
  static final String PASS = "HyperInteractive";

  public static void main(String[] args) {
    Connection conn = null;
    try {
      Class.forName(JDBC_DRIVER);
      conn = DriverManager.getConnection(DB_URL, USER, PASS);

      Statement stmt_table = conn.createStatement();
      stmt_table.execute("DROP TABLE IF EXISTS jdbc_boolean_test");
      stmt_table.execute("CREATE TABLE jdbc_boolean_test (a BOOLEAN, b INTEGER)");
      stmt_table.execute("INSERT INTO jdbc_boolean_test VALUES ('false', 1)");
      stmt_table.execute("INSERT INTO jdbc_boolean_test VALUES ('false', 2)");
      stmt_table.execute("INSERT INTO jdbc_boolean_test VALUES ('true', 3)");

      String sql = "SELECT a,b FROM jdbc_boolean_test WHERE a = ?";
      PreparedStatement stmt = conn.prepareStatement(sql);
      
      stmt.setBoolean(1, false);
      ResultSet resultSet = stmt.executeQuery();
      while (resultSet.next()) {
        System.out.println(resultSet.getBoolean("a") + " - " + resultSet.getInt("b"));
      }

      stmt.setBoolean(1, true);
      resultSet = stmt.executeQuery();
      while (resultSet.next()) {
        System.out.println(resultSet.getBoolean("a") + " - " + resultSet.getInt("b"));
      }   

      conn.close();
      System.out.println("Success");
    } catch (Exception e) {
      e.printStackTrace();
    }   
  }
}
```

```
javac brokenjdbc.java && java -cp ".:/path/to/bin/mapdjdbc-4.2.0-SNAPSHOT.jar:" brokenjdbc
```

PR #284, opened 2018-10-30
Added compatibility with osx

xmnlab:
In this PR:

- Changed `/bin/bash` to `/usr/bin/env bash` to allow bash from conda. This script need new bash, osx bash default is older that bash used by linux by default.
- Changed `id` to `whoami` command line to keep compatibility with osx
- Added SYSTEMD_PATH variable

### DIFF
--- a/systemd/install_mapd_systemd.sh
+++ b/systemd/install_mapd_systemd.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 declare -A descs
 declare -A vars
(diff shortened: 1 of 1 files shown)


xmnlab:
cc @tonyfast

andrewseidl:
Thanks @xmnlab !

What do you plan on using this for? [systemd](https://en.wikipedia.org/wiki/Systemd) only exists on Linux, which is why there are a few Linux-isms in there. We might want a separate script for macOS with defaults for that platform.

xmnlab:
Oh I see ... I didn't know that ... thanks for the explanation! I just know that it doesn't work inside a docker container.

> What do you plan on using this for? systemd only exists on Linux, which is why there are a few Linux-isms in there.

Actually I was not planning to use systemd (because it doesn't work inside docker container) I just planned to change this script and allow it to run in macos environment (called from any folder) and on the conda build task I will comment via sed the line with `sudo systemctl daemon-reload`

> We might want a separate script for macOS with defaults for that platform.

I see ... do you want to keep any of these changes in this PR?

for macOS, should I create a similar file but without the systemd stuff?

simoneves:
Closing as obsolete

PR #290, opened 2018-11-20
[WIP] Executing Python based UDFs in mapd server with numba acceleration

pearu:
This PR implements a prototype of User-Defined Functions (UDFs) for `pymapd` clients.

The UDFs are defined as Python functions decorated with `mapd()` decorator (from [mapd_udf.py](https://github.com/Quansight/mapd-core/blob/udf-pearu/python/mapd_udf.py) script).
These functions can be used to construct SQL queries to mapd-core server where the corresponding numba-compiled versions of the functions are executed.

For example:
```
import pymapd
from mapd_udf import mapd

@mapd()
def foo(x):
    return x+2

c = pymapd.connect(user="mapd", password= "HyperInteractive", host="localhost", dbname="mapd").con.cursor()

query = f"SELECT {foo('some_column')} FROM some_table"   # assuming Python version >= 3.6
c.execute(query)
# list(c) will contain the values of an expression `some_column+2`.
```
See [example1.py](https://github.com/Quansight/mapd-core/blob/udf-pearu/python/example1.py) for more examples and how to prepare mapd server to enable UDFs support. The mapd-core must be built with the following cmake options: `-DENABLE_EMBED_PYTHON=on -DENABLE_CUDA=off -DPREFER_STATIC_LIBS=off`.

*Warning*: this implementation of UDFs should be considered as a prototype and there is a number of implementation details that needs to be improved before considering it production ready. Also, many implementation choices are not optimal due to limited features of mapd-core extension functions. For instance, implementing non-scalar arguments support of mapd-core extension functions would allow a more straightforward implementation of Python based UDFs support.


asuhan:
Very nice! How hard would it be to decouple this from `pymapd`? For example, have a server endpoint which takes a Python function in text format and makes it available as an UDF. That way, any client could define UDFs.

pearu:
Let me also add that it is convenient to define UDFs within Python, especially, when the arguments have different types. By default, it is assumed that the arguments are floats and the result of the function will be a float as well. To specify other types for arguments or return value, one can use Python annotations. For instance:
```
@mapd()
def foo(a: int, b: float) -> int:
    return a + int(b)
```
Currently, the supported type annotations can be Python scalar types of types defined in `numba.types`. Support for other type annotations (eg numpy dtypes, etc) can be added as long as these can be related to `numba.types` (because in the end, we process the function via `numba`).

pearu:
Replaced with https://github.com/omnisci/mapd-core-internal/pull/3246

Issue #315, opened 2019-01-24
/usr/bin/java hardcoded in mapd-server conflicts with conda installation


pearu:
Currently, `/usr/bin/java` is hardcoded in

https://github.com/omnisci/mapd-core/blob/568e77d9c9706049eeccd32846387f4e042588d0/Calcite/Calcite.cpp#L95

that cause problems when using mapd-core from conda environment.
If `/usr/bin/java` is not installed (as conda provides java) then the server crashes with segfault.

In addition, if the conda provided java and /usr/bin/java have different versions, weird things may happen.

 Suggestion: define a cmake variable where one can specify the location of the used java program.

PR #316, opened 2019-01-29
Determine java executable path at cmake configuration.

pearu:
Resolves issue #315 .

### DIFF
--- a/Calcite/Calcite.cpp
+++ b/Calcite/Calcite.cpp
@@ -51,15 +51,15 @@ namespace {
 template <typename XDEBUG_OPTION,
           typename REMOTE_DEBUG_OPTION,
           typename... REMAINING_ARGS>
-int wrapped_execl(char const* path,
-                  XDEBUG_OPTION&& x_debug,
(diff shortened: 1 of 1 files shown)


pearu:
@randyzwitch please review.

andrewseidl:
Instead of setting the path at compile time (which is problematic if you build on one machine but try to run on another), could we maybe use `execlp` instead of `execl` and then just tell it to run whatever `java` is in your `PATH`?

pearu:
@andrewseidl , I have switched to using `execlp` that determines the `java` executable at server start.

andrewseidl:
Code looks fine, I need to make sure it doesn't break anything in a few envs.

andrewseidl:
Looks good from my testing on Arch, CentOS 7, and Ubuntu 18.04. Just waiting on some comments internally before merging.

pearu:
Although, the latest mapd-core (v 4.5.0) is released after merging this PR, the PR content is not included in v 4.5.0.

PR #318, opened 2019-02-12
Fixing assertion failure when query falls back to cpu when a subquery ran in gpu mode

dpayne:
This fixes an issue where assert failures are triggered on certain queries. If a sub query runs on gpu but the outer query fails to run on gpu due to space limitations, the whole query is retried on gpu. This then causes a few checks to fail.

### DIFF
--- a/QueryEngine/RelAlgAbstractInterpreter.h
+++ b/QueryEngine/RelAlgAbstractInterpreter.h
@@ -480,6 +480,7 @@ class RelAlgNode {
   virtual ~RelAlgNode() {}
 
   void setContextData(const void* context_data) const {
+    CHECK(!context_data_);
     context_data_ = context_data;
   }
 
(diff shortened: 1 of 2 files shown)


asuhan:
@alexbaden You might want to have a look, one of the reasons we were reluctant to allow automatic retry on CPU was this sort of careful handling of each step (and the surprising performance cliff). That being said, this lgtm and it's probably at least a partial fix.

QueryEngine/RelAlgAbstractInterpreter.h
```diff
@@ -480,7 +480,6 @@ class RelAlgNode {
   virtual ~RelAlgNode() {}
 
   void setContextData(const void* context_data) const {
-    CHECK(!context_data_);
```
alexbaden: I was able to remove this in my testing and not see any asserts. Is this removal required?

dpayne: Yeah I think so, this is where it failed first but the fix in RelAlgExecutor.cpp prevents this from happening.

alexbaden: Do you mind putting the check back in place? I am a bit worried about leaking `context_data_` if it wasn't properly cleaned up (though I think right now it only comes from the `row_set_memory_owner`, which should be cleaned up, but I'm not 100% sure on that one).

Issue #124, opened 2017-10-26
addBatch INSERT not using the column order if specified in the INSERT statement


dwayneberry:
see

https://community.mapd.com/t/mapd-java-jdbc-preparedstatement-bug/693/2

When we fix this we should also convert the api call over to use `load_table_binary_columnar`

randyzwitch:
@niviksha It would be great if we could prioritize a fix for this, as it has downstream impacts with other data ingestion tools (Streamsets specifically in my case). When loading data in batch, JDBC doesn't respect the column order, instead alphabetizing the columns. This is a pretty odd limitation, and one that's hard to understand from an end user's perspective.

randyzwitch:
Fixed by #335

PR #324, opened 2019-02-24
Make JDBC driver respect column order for batch upload (issue #124)

cdessanti:
When you try to load data in batch the column order specified on the insert statement is respected.

I placed the various checks and logic on the addBatch method to be able to throw SQLExceptions.

As Micheal said it would be better to use the load_table_binary call as it would be probably faster, but I preferred to adopt a conservative path.

### DIFF
--- a/java/omniscijdbc/src/main/java/com/omnisci/jdbc/OmniSciPreparedStatement.java
+++ b/java/omniscijdbc/src/main/java/com/omnisci/jdbc/OmniSciPreparedStatement.java
@@ -355,9 +355,9 @@ class OmniSciPreparedStatement implements PreparedStatement {
         Matcher matcher = REGEX_LOF_PATTERN.matcher(currentSQL);
         if (matcher.find()) {
           String listOfFields[] = matcher.group(1).trim().split("\\s*,+\\s*,*\\s*");
-          if (listOfFieds.length != parmCount) throw new SQLException("Exception: too many or too few values");
-          else if (Arrays.stream(listOfFieds).distinct().toArray().length != listOfFieds.length) throw new SQLException("Exception: duplicated column name");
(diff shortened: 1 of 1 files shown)


java/omniscijdbc/src/main/java/com/omnisci/jdbc/OmniSciPreparedStatement.java
```diff
@@ -346,17 +351,41 @@ public void addBatch() throws SQLException {
     if (isInsert) {
       // take the values and use stream inserter to add them
       if (isNewBatch) {
+        // check for columns names 
+        Matcher matcher = REGEX_LOF_PATTERN.matcher(currentSQL);
+        if (matcher.find()) {
+          String listOfFieds[] = matcher.group(1).trim().split("\\s*,+\\s*,*\\s*");
```
randyzwitch: Can you switch this to say `listOfFields` (i.e. fix the misspelling of `Fieds`) ?

randyzwitch:
@jack-mapd Would you be able to review this patch to the JDBC driver?

java/omniscijdbc/src/main/java/com/omnisci/jdbc/OmniSciPreparedStatement.java
cdessanti: did a commit to correct the misspelling

randyzwitch:
@cdessanti Can you revert your last commit 'Correction of issues' and move it to a separate pull request? The more issues you try and tackle in one pull request, the harder it is to review.

dwayneberry:
This change has been rolled up into https://github.com/omnisci/mapd-core/pull/335
Will close this PR

PR #325, opened 2019-02-24
Comment out needs : cxx11

### LINKED ISSUE #321: Calling needs :cxx11 is disabled! There is no replacement
Macos version: 10.14.2 (18C54)

Homebrew 2.0.1-41-gff898bc
Homebrew/homebrew-core (git revision c0b41; last commit 2019-02-13)
Homebrew/homebrew-cask (git revision d9580; last commit 2019-02-13)

''mapd-deps-osx.sh" in the scripts folder

brew install -s ../ThirdParty/Arrow/apache-arrow.rb 
Error: Calling needs :cxx11 is disabled! There is no replacement.

randyzwitch:
Per https://github.com/OSGeo/homebrew-osgeo4mac/issues/626 and others, 
Homebrew changed its compiler toolchain such that this line causes 
errors. Removing this line allows for installation to continue.

Fixes #321

### DIFF
--- a/ThirdParty/Arrow/apache-arrow.rb
+++ b/ThirdParty/Arrow/apache-arrow.rb
@@ -11,7 +11,7 @@ class ApacheArrow < Formula
   depends_on "python" => :optional
   depends_on "python@2" => :optional
 
-  needs :cxx11
+  # needs :cxx11
 
   def install
     ENV.cxx11


randyzwitch:
This appears to work, but I can only confirm for Mojave. @xmnlab are you using Sierra or High Sierra?

andrewseidl:
@fleapapa could you see if the upstream formula works for you now? Looks like you started vendoring it due to some issues with Parquet and Snappy.

https://github.com/Homebrew/homebrew-core/blob/master/Formula/apache-arrow.rb

fleapapa:
We needed the vendored .rb because original factory apache-arrow.rb didn't include Parquet.

Will try this new rb. Just suspecting that it enables Snappy...

randyzwitch:
Let's merge this. We can always go back and unvendor Arrow if its no longer needed here. Otherwise, this fix allows for compiling on Mojave, which is the most important thing IMO.

PR #328, opened 2019-02-26
Add unit test for punt to CPU with subquery

alexbaden:
Added a quick unit test to repro the issue fixed by #318.

@dpayne Would you mind taking a look?


dpayne:
That should trigger it. My specific query was a CTE e.g. `WITH some_table ( ... ) ...`, but this should also trigger it.

PR #332, opened 2019-03-05
Fixing declaration of char[] buffers for csv import

dpayne:
This came up when recompiling with jemalloc. The database hung while importing from csv. I then found the issue with the address sanitizer.


asuhan:
I think it's off because of the generated bison code. Yet another reason to look into replacing that.

alexbaden:
Hrmm, maybe -- the setting has been in there since I've been paying attention to ASAN. But, I was able to get it to pass with GCC just by cleaning up some other problem areas (NoneEncoder, PopulateTableRandom).

PR #367, opened 2019-07-23
Support JIT code profiling using Intel VTune.

ienkovich:


ienkovich:
I merged with master branch and applied clang-format. Sanity tests don't show any regressions. Do I need to make some additional testing?

PR #370, opened 2019-08-02
Add LICM pass and enable it for DateAdd call

ienkovich:
Running TPC-H benchmark queries on Omnisci we found that for some queries generated row_func functions have datetime computations (using DateAdd function call) on literals. When row_func is inlined these computations become loop invariant but are not moved out from the loop. In some cases DateAdd call leads to gmtime_r_newlib call which is quite expensive.

This patch add Loop Invariant Code Motion pass to pass manager and add function attributes for DateAdd declaration to correctly identify it as a loop invariant.

This patch gives ~3x speed-up for TPC-H queries Q5, Q10 and ~20x speed-up for Q6, Q12. Actually I could only run 11 of 22 queries of the benchmark so other queries also might be affected.

LICM pass seems to be quite useful and not expensive (it should be linear to total number of statements in loops), so I propose to always use it.

### DIFF
--- a/QueryEngine/CgenState.h
+++ b/QueryEngine/CgenState.h
@@ -199,15 +199,27 @@ struct CgenState {
   }
   // look up a runtime function based on the name, return type and type of
   // the arguments and call it; x64 only, don't call from GPU codegen
-  llvm::Value* emitExternalCall(const std::string& fname,
-                                llvm::Type* ret_type,
-                                const std::vector<llvm::Value*> args) {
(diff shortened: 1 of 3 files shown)


asuhan:
We used to have LICM for hoisting literal loads, glad it's making such a strong comeback. Thanks, I'll have a look later today.

cdessanti:
@ienkovich,

Amazing job. I also noticed that looking at llvm code generated by optimizer

PR #372, opened 2019-08-06
Use arithmetic with overflow intrinsics to detect overflows

ienkovich:
This patch utilizes arithmetic with overflow LLVM intrinsics to detect overflows/underflows. On x86 CPU this allows to use overflow bit flag instead of doing reverse operations for the check. This is especially useful for multiplication case. TPC-H query Q1 has multiplication with overflow check and this patch gives it 6.4x speed-up.

I'm not sure about implications for CUDA. Used intrinsics are not x86 specific so I expect them to work on all platforms. Unfortunately I don't have CUDA on my server to test it.

### DIFF
--- a/QueryEngine/ArithmeticIR.cpp
+++ b/QueryEngine/ArithmeticIR.cpp
@@ -630,7 +630,8 @@ llvm::Value* CodeGenerator::codegenBinOpWithOverflowCheck(
   // In case of null check we have to use NULL result on check fail
   if (null_check) {
     auto phi = cgen_state_->ir_builder_.CreatePHI(ret->getType(), 2);
-    phi->addIncoming(llvm::ConstantInt::get(ret->getType(), inline_int_null_val(ti)), null_check);
+    phi->addIncoming(llvm::ConstantInt::get(ret->getType(), inline_int_null_val(ti)),
+                     null_check);
     phi->addIncoming(ret, val_bb);
     ret = phi;
   }
--- a/QueryEngine/ArithmeticIR.cpp
+++ b/QueryEngine/ArithmeticIR.cpp
@@ -61,7 +61,7 @@ llvm::Value* CodeGenerator::codegenArith(const Analyzer::BinOper* bin_oper,
     CHECK_EQ(lhs_type.get_type(), rhs_type.get_type());
   }
   if (lhs_type.is_integer() || lhs_type.is_decimal() || lhs_type.is_timeinterval()) {
-    return codegenIntArith(bin_oper, lhs_lv, rhs_lv);
+    return codegenIntArith(bin_oper, lhs_lv, rhs_lv, co);
   }
   if (lhs_type.is_fp()) {
     return codegenFpArith(bin_oper, lhs_lv, rhs_lv);
@@ -73,7 +73,8 @@ llvm::Value* CodeGenerator::codegenArith(const Analyzer::BinOper* bin_oper,
 // Handle integer or integer-like (decimal, time, date) operand types.
 llvm::Value* CodeGenerator::codegenIntArith(const Analyzer::BinOper* bin_oper,
                                             llvm::Value* lhs_lv,
-                                            llvm::Value* rhs_lv) {
+                                            llvm::Value* rhs_lv,
+                                            const CompilationOptions& co) {
   const auto lhs = bin_oper->get_left_operand();
   const auto rhs = bin_oper->get_right_operand();
   const auto& lhs_type = lhs->get_type_info();
@@ -88,21 +89,24 @@ llvm::Value* CodeGenerator::codegenIntArith(const Analyzer::BinOper* bin_oper,
                         rhs_lv,
                         null_check_suffix.empty() ? "" : int_typename,
                         null_check_suffix,
-                        oper_type);
+                        oper_type,
+                        co);
     case kPLUS:
       return codegenAdd(bin_oper,
                         lhs_lv,
                         rhs_lv,
                         null_check_suffix.empty() ? "" : int_typename,
                         null_check_suffix,
-                        oper_type);
+                        oper_type,
+                        co);
     case kMULTIPLY:
       return codegenMul(bin_oper,
                         lhs_lv,
                         rhs_lv,
                         null_check_suffix.empty() ? "" : int_typename,
                         null_check_suffix,
-                        oper_type);
+                        oper_type,
+                        co);
     case kDIVIDE:
       return codegenDiv(lhs_lv,
                         rhs_lv,
@@ -209,7 +213,8 @@ llvm::Value* CodeGenerator::codegenAdd(const Analyzer::BinOper* bin_oper,
                                        llvm::Value* rhs_lv,
                                        const std::string& null_typename,
                                        const std::string& null_check_suffix,
-                                       const SQLTypeInfo& ti) {
+                                       const SQLTypeInfo& ti,
+                                       const CompilationOptions& co) {
   CHECK_EQ(lhs_lv->getType(), rhs_lv->getType());
   CHECK(ti.is_integer() || ti.is_decimal() || ti.is_timeinterval());
   llvm::Value* chosen_max{nullptr};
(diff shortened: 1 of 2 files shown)


andrewseidl:
Thanks! Let me check with our devops team to see if we can whitelist this PR to run on our public Jenkins instance (which has GPUs).

In the meantime, unfortunately it looks like it is failing on GPU:
```
[ RUN      ] CodeGeneratorTest.IntegerExprGPU
2019-08-06T16:01:13.649552 F 22113 NativeCodegen.cpp:828 use of undefined value '@llvm.sadd.with.overflow.i32'
```
```
[ RUN      ] Select.FilterAndSimpleAggregation
2019-08-06T15:21:03.139855 F 1741 NativeCodegen.cpp:828 use of undefined value '@llvm.sadd.with.overflow.i64'
```

Trace:
```
2019-08-06T16:06:02.329417 F 26576 NativeCodegen.cpp:828 use of undefined value '@llvm.sadd.with.overflow.i32'                                                
                                                                               
Thread 1 "CodeGeneratorTe" received signal SIGABRT, Aborted.                                                                                                  
0x00007fffedecb755 in raise () from /usr/lib/libc.so.6                   
(gdb) bt                                                                                                                                                      
#0  0x00007fffedecb755 in raise () from /usr/lib/libc.so.6                                                                                                                                                                                                                                                                  
#1  0x00007fffedeb6851 in abort () from /usr/lib/libc.so.6                                                                                                                                                                                                                                                                  
#2  0x0000555557243f1e in logger::Logger<logger::Severity, (logger::Severity)7>::~Logger (this=0x7fffffffd400, __in_chrg=<optimized out>)                     
    at ../Shared/Logger.cpp:236                                                
#3  0x000055555774a5fd in CodeGenerator::generatePTX (                                                                                                                                                                                                                                                                      
    cuda_llir="\ndeclare void @llvm.dbg.declare(metadata, metadata, metadata)\ndeclare void @llvm.dbg.value(metadata, metadata, metadata)\ndeclare void @llvm.lifetime.start(i64, i8* nocapture) nounwind\ndeclare void @ll"..., nvptx_target_machine=0x55555845e500, cgen_state=0x555558494410)
    at ../QueryEngine/NativeCodegen.cpp:828                             
#4  0x00005555577491ba in CodeGenerator::generateNativeGPUCode (func=0x5555584ac188, wrapper_func=0x555558113948,                                             
    live_funcs=std::unordered_set with 2 elements = {...}, co=..., gpu_target=...) at ../QueryEngine/NativeCodegen.cpp:724                                    
#5  0x0000555557ac860b in ScalarCodeGenerator::generateNativeGPUCode (this=0x7fffffffe100, func=0x5555584ac188, wrapper_func=0x555558113948, co=...)          
    at ../QueryEngine/ScalarCodeGenerator.cpp:189                         
#6  0x0000555557ac8087 in ScalarCodeGenerator::generateNativeCode (this=0x7fffffffe100, compiled_expression=..., co=...)                                      
    at ../QueryEngine/ScalarCodeGenerator.cpp:151
#7  0x00005555571ce72c in CodeGeneratorTest_IntegerExprGPU_Test::TestBody (this=0x555558a03820) at ../Tests/CodeGeneratorTest.cpp:346                         
#8  0x000055555722c81b in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (object=0x555558a03820,                              
    method=&virtual testing::Test::TestBody(), location=0x55555675faf3 "the test body") at ../ThirdParty/googletest/gmock-gtest-all.cc:3899                   
#9  0x00005555572244dd in testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=0x555558a03820,                                 
    method=&virtual testing::Test::TestBody(), location=0x55555675faf3 "the test body") at ../ThirdParty/googletest/gmock-gtest-all.cc:3935                   
#10 0x00005555571f977e in testing::Test::Run (this=0x555558a03820) at ../ThirdParty/googletest/gmock-gtest-all.cc:3973                                        
#11 0x00005555571fa0d5 in testing::TestInfo::Run (this=0x5555580d2870) at ../ThirdParty/googletest/gmock-gtest-all.cc:4149                                    
#12 0x00005555571fa754 in testing::TestCase::Run (this=0x5555580d1d70) at ../ThirdParty/googletest/gmock-gtest-all.cc:4267                                    
#13 0x00005555572054cb in testing::internal::UnitTestImpl::RunAllTests (this=0x5555580d1980) at ../ThirdParty/googletest/gmock-gtest-all.cc:6633              
#14 0x000055555722d701 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x5555580d1980,            
    method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x555557205226 <testing::internal::UnitTestImpl::RunAllTests()>, location=0x5555567604d0 "auxiliary test code (environments or event listeners)") at ../ThirdParty/googletest/gmock-gtest-all.cc:3899                       
#15 0x0000555557225471 in testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x5555580d1980,               
    method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x555557205226 <testing::internal::UnitTestImpl::RunAllTests()>, location=0x5555567604d0 "auxiliary test code (environments or event listeners)") at ../ThirdParty/googletest/gmock-gtest-all.cc:3935                       
#16 0x0000555557203f7a in testing::UnitTest::Run (this=0x555558067920 <testing::UnitTest::GetInstance()::instance>)                                           
    at ../ThirdParty/googletest/gmock-gtest-all.cc:6242                        
#17 0x00005555571d10b0 in RUN_ALL_TESTS () at ../ThirdParty/googletest/gtest/gtest.h:22104         
#18 0x00005555571ceed9 in main (argc=1, argv=0x7fffffffe688) at ../Tests/CodeGeneratorTest.cpp:390                                                            
```

Also seeing a warning, at least with GCC 9.1.0 / LLVM 8.0.1, need to see if this was existing or not:
```
../QueryEngine/ArithmeticIR.cpp: In member function ‘llvm::Value* CodeGenerator::codegenBinOpWithOverflowCheck(const Analyzer::BinOper*, llvm::Value*, llvm::Value*, const string&, const SQLTypeInfo&)’:
../QueryEngine/ArithmeticIR.cpp:616:38: warning: ‘fn_id’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  616 |       llvm::Intrinsic::getDeclaration(cgen_state_->module_, fn_id, {lhs_lv->getType()});
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

mattdlh:
I've added @ienkovich to the contributors whitelist. Going forward, new commits to this PR will trigger a  build of https://jenkins-os.mapd.com/job/core-pr

shtilman:
Nice proposal!
Unfortunately, we have to keep our codegen pretty generic to be able to push the same IR through to both CPU and GPU.

The set of intrinsics that ptx backend would accept is quite limited, overflow intrinsics are not in it:
https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#intrinsic-functions

Perhaps we could just specialize CPU codegen path to take advantage of these.

asuhan:
Yes, let's route to GPUs and CPUs to use different lowerings here, the speedup is worth it.

QueryEngine/ArithmeticIR.cpp
```diff
@@ -660,3 +572,69 @@ llvm::Value* CodeGenerator::codegenUMinus(const Analyzer::UOper* uoper,
...
+    case kPLUS:
+      bb_prefix = "add";
+      fn_id = llvm::Intrinsic::sadd_with_overflow;
+      break;
+    case kMULTIPLY:
+      bb_prefix = "mul";
+      fn_id = llvm::Intrinsic::smul_with_overflow;
+      break;
+    default:
+      CHECK(false);
```
asuhan: We're gradually replacing instances of `CHECK(false)` with `LOG(FATAL)` with a meaningful error message. Also, you might need to initialize `fn_id` if we can't make the compiler see that the path through `default` is unreachable, to squash the warning. Maybe move this `switch` to a helper function as well, it's easy to isolate and makes this method smaller.

asuhan:
@ienkovich Is avoiding the division the main reason for the speedup? The additional comparison we do besides that shouldn't weigh too much, but division is dozens of cycles iirc.

QueryEngine/ArithmeticIR.cpp
ienkovich: Tanks for comments! I'll rework this part.

ienkovich:
> @ienkovich Is avoiding the division the main reason for the speedup? The additional comparison we do besides that shouldn't weigh too much, but division is dozens of cycles iirc.

Right. Profile for query shows we spend most of time in divisions. Two additional divisions for a single multiply become a huge overhead.

ienkovich:
> Yes, let's route to GPUs and CPUs to use different lowerings here, the speedup is worth it.

OK. I'll rework the patch.

QueryEngine/CodeGenerator.h
```diff
@@ -406,6 +412,15 @@ class CodeGenerator {
       const std::unordered_map<llvm::Value*, llvm::Value*>&,
       const CompilationOptions&);
 
+  llvm::Function* getArithWithOverflowFunction(const Analyzer::BinOper* bin_oper,
```
asuhan: I would add comments to these two methods to describe their functionality. Also, the `getArithWithOverflowFunction` name confused me a bit, maybe call it `getArithWithOverflowIntrinsic`.

ienkovich: Done!

asuhan:
@andrewseidl Let's get a full test run for this, if it looks good we should merge this.

PR #378, opened 2019-08-19
Import/Importer: Use string_view instead on copying strings in csv reader

fexolm:
I've implemented initial support for string_view in Importer.
I haven't optimized all possible branches, like double conversions yet.

This implementation shows almost 2x speedup when importing ny-taxi benchmark with a single thread.


Import/Importer.cpp
```diff
@@ -270,35 +290,34 @@ static const char* get_row(const char* buf,
...
+          auto s = trim_space<String>(field, p - field);
           row.push_back(s);
         } else {
-          auto field_buf = std::make_unique<char[]>(p - field + 1);
           int j = 0, i = 0;
           for (; i < p - field; i++, j++) {
             if (has_escape && field[i] == copy_params.escape &&
                 field[i + 1] == copy_params.quote) {
-              field_buf[j] = copy_params.quote;
+              field[j] = copy_params.quote;
```
msariyuce: You are modifying the given buffer (field( = buf )) instead of creating another buffer, which makes sense if we are sure that it will not used again. However if there is a problem in multithreaded calls, this buffer will be used from the beginning with a single thread. So I am guessing that the buffer must not be modified.

Shared/StringToDatum.h
```diff
@@ -0,0 +1,268 @@
...
+
+#include <Utils/StringConversions.h>
+#include <boost/utility/string_view.hpp>
+#include <cinttypes>
+#include "../Shared/StringToDatum.h"
+#include "DateConverters.h"
+#include "TimeGM.h"
+#include "sqltypes.h"
+
+template <typename String>
```
msariyuce: Would it be better if we keep the implementations in cpp file?
To avoid compile errors, one can include cpp file at the end of the header file.

Utils/StringConversions.h
```diff
@@ -0,0 +1,63 @@
...
+const inline std::string& to_string(const std::string& s) {
+  return s;
+}
+
+std::string inline to_string(const boost::string_view& s) {
+  return s.to_string();
+}
+
+template <typename String>
+int64_t strtol(const String& s) {
```
msariyuce: I prefer `std::stoll(s.to_string())` instead of writing the function from scratch.

fexolm: It is one of the main optimizations here. `std::stoll(to_string(s))` will cause copying, which is unnecessary here. I could use c++17 `std::to_chars` since you are moving to c++17. But I'm not sure how it implemented. Will need some benchmarking here.

Shared/StringToDatum.h
fexolm: You mean creating StringToDatum.cpp and include it at the end of StringToDatum.h?

Import/Importer.cpp
fexolm: AFAIK. Buffer is used just within a single thread, so it wouldn't cause a problem.

msariyuce: > AFAIK. Buffer is used just within a single thread, so it wouldn't cause a problem.

Could you please check this function `Detector::split_raw_data()` ?

Shared/StringToDatum.h
msariyuce: Yes

PR #398, opened 2019-09-13
Add flag to pass LVVM options

ienkovich:
Hi!

Working on IR analysis and optimizations I found very useful flag to dump all generated modules to files. 'Explain' feature is very handy but sometimes its result differs from actually used module and also there are queries which require multiple modules. Using new debug dump flag you can get all module IRs (original and optimized) in tmp files. I just want to share this patch in case someone also needed similar feature.

Another added flag allows to pass additional flags to LLVM. For example this can be used to pass debug print options and get module dumps after each optimization pass.

I remember someone mentioned new debug features are under development. So feel free to ignore this request if something better is on its way.


ienkovich:
Thanks for the log channels tip! I'll add a commit to keep LLVM opts flag only.

alexbaden:
Hi @ienkovich -- sorry it has taken so long for us to get back to you. Could you please rebase and fix the clang-tidy failure? Then we'll get this in.

anmyachev:
@alexbaden ping

Issue #392, opened 2019-09-10
[Perf] Explpore the influence of LLVM passes on the benchmarks


aregm:

psilvest:
This issue has been inactive for a while, and the project has changed substantially. If this still occurs on the latest release, please add reproduction steps and environment details.

PR #399, opened 2019-09-16
Added possibility to compile jit code with llvm's optimization levels O1-O3

Garra1980:
This is an attempt to allow choosing llvm's optimization level for compiling JIT code. In some case this could give good performance boost (20-30% on some tpc-h queries)

Targeting #392

### DIFF
--- a/QueryEngine/CompilationOptions.h
+++ b/QueryEngine/CompilationOptions.h
@@ -19,7 +19,7 @@
 
 enum class ExecutorDeviceType { CPU, GPU };
 
-enum class ExecutorOptLevel { Default, LoopStrengthReduction, ReductionJIT };
+enum class ExecutorOptLevel { Default, O1, O2, O3, LoopStrengthReduction, ReductionJIT };
 
 enum class ExecutorExplainType { Default, Optimized };
 
--- a/QueryEngine/CompilationOptions.h
+++ b/QueryEngine/CompilationOptions.h
@@ -19,7 +19,14 @@
 
 enum class ExecutorDeviceType { CPU, GPU };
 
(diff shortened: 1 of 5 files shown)


QueryEngine/CompilationOptions.h
```diff
@@ -19,7 +19,7 @@
 
 enum class ExecutorDeviceType { CPU, GPU };
 
-enum class ExecutorOptLevel { Default, LoopStrengthReduction, ReductionJIT };
+enum class ExecutorOptLevel { Default, O1, O2, O3, LoopStrengthReduction, ReductionJIT };
```
shtilman: Need a comment saying that O1, O2, O3 shouldn't move.
Or should assign values explicitly O1=1, O2=2, O3=3

QueryEngine/NativeCodegen.cpp
```diff
@@ -100,23 +101,34 @@ void optimize_ir(llvm::Function* query_func,
                  llvm::Module* module,
                  const std::unordered_set<llvm::Function*>& live_funcs,
                  const CompilationOptions& co) {
-  llvm::legacy::PassManager pass_manager;
+  if (co.opt_level_ >= ExecutorOptLevel::O1 && co.opt_level_ <= ExecutorOptLevel::O3) {
```
shtilman: Could we pick a consistent naming theme in this function? mpm, pass_manager..

QueryEngine/CompilationOptions.h
Garra1980: Done

QueryEngine/NativeCodegen.cpp
Garra1980: Done

PR #414, opened 2019-10-10
Sphinx docs - add analytics

mattdlh:
GitHub actions will comment out the line including the segment .js file, so it will only report analytics when build in GH actions on master before getting hosted from `gh-pages` branch.


mattdlh:
retest this please

PR #417, opened 2019-10-17
fix doc typos

ienkovich:


docs/source/execution/overview.rst
```diff
@@ -38,7 +38,7 @@ Apache Calcite
 OmniSciDB uses Apache Calcite for frontline query parsing and cost-based optimization. Calcite runs as a separate process and communicates with the server using an Apache Thrift :doc:`../data_model/api`. Calcite returns an optimized relational algebra tree serialized via JSON. The serialized JSON string is passed to the ``RelAlgExecutor`` for execution.
 
 .. note::
-    Calcite is used only for DDL statements. For DML statements,
+    Calcite is used only for DML statements. For DDL statements,
```
alexbaden: This is at least more correct -- we don't use Calcite for `INSERT` (yet). The original statement was very wrong -- thanks for this!

PR #421, opened 2019-10-24
remove dead code

ienkovich:


alexbaden:
Are you sure this can't be triggered with a loop join? Admittedly the logic is a bit weird (by definition the join condition should force the column to be non-lazy -- perhaps that's the objective here?), but I'm curious how you determined this to be dead.

ienkovich:
I just noticed that we have ra_exe_unit.join_quals.empty() as a part of removed condition and it's also checked earlier. So A || (A && B) == A.

PR #449, opened 2019-11-14
run_benchmark_import.py script improved to allow importing multiple data files and specify fragment size when table is created

gshimansky:


mattdlh:
Hey @gshimansky thanks for the PR. I agree adding wildcard and custom fragment size to the import script is needed.

For the wildcard import via python `glob`; it appears that this breaks the ability to run this script against a remote system, as python is searching for the files on the local filesystem. We currently support running this script against a remote database for the `run_benchmark` and this import script.

Currently the dependency on `glob` breaks even single-file import on a remote machine, which would need to be resolved. One path would be to at least check that the machine the queries are running against matches "localhost" or somehow verify those files are available if parameters are passed in to import multiple files, and throw an error if not.

However, after some internal discussion, it came up that the db has no issue importing files using wildcard in the import statement syntax. I believe just passing the filename with wildcard into the import syntax would solve both issues - be able to import a set of files while also being able to run on a remote machine.

gshimansky:
You are right, omnisci server apparently supports wildcards inside of COPY statement. Now just star wildcards `*`, but also curly braces work for me, so it is possible to limit the number of data files using wildcard like `trips_xa{b,c,d,e,f,g,h,i,j,k}.csv.gz`.
I rebased my pull request to exclude wildcards commit, now it contains only addition for fragment size and flights schema for data available form https://data.mapd.com/flights_2008_7M.tar.gz .

mattdlh:
Great, good to hear that option works for wildcards.

Just tested the fragment size changes and they work as-is. Two small items though:

1. Can you share the logic for checking if `fragment_size` is None here:
https://github.com/omnisci/omniscidb/blob/7d17aedf0cdb2de33d8c995e4efdd00057bf4b4d/Benchmarks/run_benchmark_import.py#L373
considering the default value of `32000000`?

2. Can you change the `--fragment_size` flag from underscore to dash (`--fragment-size`) for the input flag here
https://github.com/omnisci/omniscidb/blob/7d17aedf0cdb2de33d8c995e4efdd00057bf4b4d/Benchmarks/run_benchmark_import.py#L240
to match the others?

gshimansky:
You are right, #1 was redundant, #2 fixed.

PR #450, opened 2019-11-18
Add NY Taxi Benchmark SQL schema and queries

gshimansky:
Benchmark for NY Taxi dataset. Requires changes for fragment size from #449.


mattdlh:
@gshimansky we actually have these queries internally but held on moving them to the OS repo due to questions about the taxis dataset availability. I just got the all clear to move these queries from our internal repo though. They are mostly the same with slightly different names/formatting, plus a fifth "double modulo" query.
Could I go ahead and update your PR with those slight changes and the queries we are using internally?

gshimansky:
Sure, I don't really care who writes these queries as long as they are the same :)
I just need to have ##FRAGMENT_SIZE## for table schema, which I added in #449.

mattdlh:
Updated the naming for the queries/schema and added links to the public datasets to the benchmarks README. If you are OK with the changes, this should be good to go pending required checks.

gshimansky:
Yes, changes look fine to me. Go on and merge them.

PR #460, opened 2019-12-06
omnisql: clarify \c command help text

andrewseidl:
Clarify that the `\c` command is for switching databases as a different user.

@gshimansky pointed out that since we have a separate `\db` command for switching between databases (albeit as the same user), it would seem like `\c` might instead stand for `create`.

We should probably find a better way to combine these two commands.

### DIFF
--- a/SQLFrontend/CommandFunctors.h
+++ b/SQLFrontend/CommandFunctors.h
@@ -215,7 +215,8 @@ StandardCommand(Help, {
   std::cout << "\\t [regex] List all tables, optionally matching regex.\n";
   std::cout << "\\v [regex] List all views, optionally matching regex.\n";
   std::cout << "\\d <table> List all columns of a table or a view.\n";
-  std::cout << "\\c <database> <user> <password>.\n";
+  std::cout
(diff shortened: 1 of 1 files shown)


SQLFrontend/CommandFunctors.h
```diff
@@ -215,7 +215,8 @@ StandardCommand(Help, {
   std::cout << "\\t [regex] List all tables, optionally matching regex.\n";
   std::cout << "\\v [regex] List all views, optionally matching regex.\n";
   std::cout << "\\d <table> List all columns of a table or a view.\n";
-  std::cout << "\\c <database> <user> <password>.\n";
+  std::cout
+      << "\\c <database> <user> <password> Connect to database as different user.\n";
```
wamsiv: Maybe we can use `\connect` to make it more descriptive?

gshimansky: When you explained to me that `\c` does the switch, I thought that it means `change` until I saw this pull request :)

shawnyar999: `\connect` and `\c` both are fine with me with the new documentation.

PR #494, opened 2020-01-14
Fix arrow TaskGroup TSAN warnings

fexolm:


alexbaden:
This turned out to be a false positive. We built a docker container with our deps in it compiled against TSAN, and got a pass: https://hub.docker.com/r/omnisci/core-build-tsan-centos7-cuda10.0

PR #511, opened 2020-02-10
Fix default value of n for CHAR, VARCHAR

prutskov:


andrewseidl:
Could you clarify a bit on the desired behavior/use? Though I agree this should probably get changed - elsewhere in the parser we're only throwing if it's non-negative, not caring if it's zero.

prutskov:
@andrewseidl I try to use CHAR/VARCHAR without parameters (length). This is supported by some SQL-servers. What kind of behavior do we expect using this(and other) types without parameters?

PR #535, opened 2020-03-25
Change throwng of exceptions in FSI

prutskov:


DataMgr/ForeignStorage/ArrowCsvForeignStorage.cpp
```diff
@@ -157,7 +157,8 @@ static std::shared_ptr<arrow::DataType> getArrowImportType(const SQLTypeInfo typ
       case 8:
         return int64();
       default:
-        CHECK(false);
+        throw std::runtime_error("Integer type of size " +
```
alexbaden: This one is fine, but the other two should remain fatal as their behavior is undefined. If we're seeing these methods get called we have a bug higher up the stack.

aregm:
@prutskov Can you please rebase and ensure that formatting passes?

prutskov:
@aregm This was merged already in other PR. I close this PR.

PR #537, opened 2020-03-26
add a flag to disable lazy fetch

ienkovich:
We see some significant gain on disabling lazy fetch for benchmarks we run on CPU. E. g. for Mortgage it gives 1.3x speed-up. For now we disable it using the flag but probably it should be disabled for CPU by default. We should collect more numbers to decide.


alexbaden:
Merged w/ 0e0464986d9c1688630bfd7539d09bf61b36e710

PR #539, opened 2020-03-27
FSI: Support fragment size argument

fexolm:
This pr doesn't provide support to `maxChunkSize`. I'll think about it and send separate pr, if you don't mind.  This pr is already overcomplicated I think.


fexolm:
Please review, but let's merge it after my patch to StringDictionary (I would close that and open another one, as it would change significantly with the StringDictionary patch).

PR #543, opened 2020-03-30
avoid unnecessary init and copies

ienkovich:
This patch avoids unnecessary memory zeroing for projections. Also introduces in-place buffer initialization in case of a single buffer to avoid memcpy.


alexbaden:
Merged: https://github.com/omnisci/omniscidb/commit/c97a66e209cfcd9fa00ce06e28b36ebafa245959

PR #548, opened 2020-03-31
add efficient arrow converter for numeric columnar columns

ienkovich:
Here we add arrow conversion for columnar result sets. It doesn't use array builders and therefore avoids unnecessary data copies (in case of a single result set storage a zero copy conversion is used). For now it's for numeric columns only.


jclay:
This was merged in as part of #4871.

PR #549, opened 2020-03-31
avoid full scan in ResultSet::rowCount when possible

ienkovich:
This patch speeds up row count computation for projections using a binary search instead of a linear scan.

### DIFF
--- a/QueryEngine/ResultSetIteration.cpp
+++ b/QueryEngine/ResultSetIteration.cpp
@@ -2176,7 +2176,7 @@ bool ResultSetStorage::isEmptyEntryColumnar(const size_t entry_idx,
 namespace {
 
 template <typename T>
-inline size_t makeBinSearch(size_t l, size_t r, T&& is_empty_fn) {
+inline size_t make_bin_search(size_t l, size_t r, T&& is_empty_fn) {
   // Avoid search if there are no empty keys.
   if (!is_empty_fn(r - 1)) {
     return r;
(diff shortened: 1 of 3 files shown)


QueryEngine/ResultSetIteration.cpp
```diff
@@ -2173,6 +2173,50 @@ bool ResultSetStorage::isEmptyEntryColumnar(const size_t entry_idx,
   return false;
 }
 
+namespace {
+
+template <typename T>
+inline size_t makeBinSearch(size_t l, size_t r, T&& is_empty_fn) {
```
sashkiani: Since this is a free function, can you please change its name to be `make_bin_search` so that it matches our coding style guide?

QueryEngine/ResultSetIteration.cpp
```diff
@@ -2173,6 +2173,50 @@ bool ResultSetStorage::isEmptyEntryColumnar(const size_t entry_idx,
...
+  }
+
+  return r;
+}
+
+}  // namespace
+
+size_t ResultSetStorage::binSearchRowCount() const {
+  CHECK(query_mem_desc_.getQueryDescriptionType() == QueryDescriptionType::Projection);
+  CHECK_EQ(query_mem_desc_.getEffectiveKeyWidth(), 8);
```
sashkiani: Can you please change this to be using `size_t(8)` instead? It currently produces a warning because of type mismatch in comparison.

PR #555, opened 2020-04-21
FSI optimization patches

fexolm:
- Optimize string dictionary building
- Add 'skip rows', 'has header' and 'delimiter' options for create dataframe
- Add support for decimals
- Add support for custom fragment sizes
- Move all threading to TBB


fexolm:
@alexbaden PTAL

DataMgr/ForeignStorage/ArrowCsvForeignStorage.cpp
```diff
@@ -124,19 +161,32 @@ void ArrowCsvForeignStorage::read(const ChunkKey& chunk_key,
...
       }
     } else {
       // TODO: nullify?
       auto fixed_type = dynamic_cast<arrow::FixedWidthType*>(array_data->type.get());
       if (fixed_type) {
-        sz = array_data->length * fixed_type->bit_width() / 8;
-      } else
+        sz = size * (fixed_type->bit_width() / 8);
+      } else {
         CHECK(false);  // TODO: what's else???
```
anton-malakhov, line 185: it's still here.. can we do something with this branch?

config/tsan.suppressions
```diff
@@ -10,3 +10,5 @@ race:apache::thrift::server
 thread:libcuda.so
 # 2020-02-23: resolve thread safety issues when acquiring string dict proxy through resultset 
 race:ResultSet::makeTargetValue 
+
+race:^tbb*
```
alexbaden, line 14: Is this coming from the same place as the arrow race previously?

Catalog/TableDescriptor.h
```diff
@@ -54,6 +54,9 @@ struct TableDescriptor {
   int64_t maxChunkSize;    // max number of rows per fragment
   int32_t fragPageSize;    // page size
   int64_t maxRows;         // max number of rows in the table
+  int64_t skipRows;        // number of skipped rows of data in CSV file
```
alexbaden, line 57: I'd like to find a different place for this -- one option might be to create a `DataframeTableDescriptor` which could inherit from TD but supply the addition options for transport to FSI. The TableDescriptor API is already too big -- it should ideally contain just the base id, shard, etc for the table.

DataMgr/ForeignStorage/ArrowCsvForeignStorage.cpp
```diff
@@ -211,36 +261,255 @@ void ArrowCsvForeignStorage::prepareTable(const int db_id,
...
+      });
+}
+
 void ArrowCsvForeignStorage::registerTable(Catalog_Namespace::Catalog* catalog,
                                            std::pair<int, int> table_key,
                                            const std::string& info,
                                            const TableDescriptor& td,
                                            const std::list<ColumnDescriptor>& cols,
                                            Data_Namespace::AbstractBufferMgr* mgr) {
+  // tbb::task_scheduler_init init(1);
```
alexbaden, line 488: Commented code

DataMgr/ForeignStorage/ArrowCsvForeignStorage.cpp
fexolm: It is unreachable as we don't know any cases, that can go that way. So it's ok to paste CHECK(false) in the unreachable branch.

anton-malakhov: then, remove/reword the comment

config/tsan.suppressions
fexolm: Not really. As we found, building tbb with TSAN and ignoring race:rml::* (which is tbb internal resource manager) also takes warnings away. So, we can assume that there are no races in user code. 
The only warning remains is arrow::read_csv which is of cause false positive, as arrow is well tested with tsan enabled on their side. 
So, it seems, there are no problems on our side, but we need to somehow disable TSAN warnings.

fexolm:
@alexbaden PTAL.

Catalog/Catalog.cpp
```diff
@@ -2765,16 +2768,21 @@ void Catalog::createShardedTable(
     const list<ColumnDescriptor>& cols,
     const std::vector<Parser::SharedDictionaryDef>& shared_dict_defs) {
   cat_write_lock write_lock(this);
+  DataframeTableDescriptor* df_td = dynamic_cast<DataframeTableDescriptor*>(&td);
```
alexbaden, line 2771: This doesn't seem quite right -- can you explain?

prutskov: This's added for determining which of type of descriptor is using now. In case when `td` contains `TableDescriptor` object `df_dt` will be `nullptr` and we initialize `df_dt` with DataframeTableDescriptor(TableDescriptor) constructor.  It's needed as we create locally objects of descriptors at lines 2778 and 2785. In general, this dynamic cast may be moved to `DataframeTableDescriptor` constructor

alexbaden:
I merged a large portion of this w/ https://github.com/omnisci/omniscidb/commit/a27c6e7837d6494d7426bc823d8df7058702ddf2 

We need to revisit the computation of metadata in the encoders, but otherwise everything else should be in master now.

PR #558, opened 2020-05-20
[WIP] TBB threading layer

anton-malakhov:
Adding threading layer: asyncs are enabled on top of TBB. Almost all std::async were enabled excluding these doing I/O. a few parallel_fors were added instead of similar async patterns.
It raises a few architectural questions.
E.g. do we want threading layers to be selectable at runtime or compile-time is fine?
Do we want alternative layer to be Arrow-based, std::-based or something else?
Squashed from https://github.com/intel-go/omniscidb/pull/29 removing all the merge artifacts.

@alexbaden @fexolm @aregm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@
 *.out
 *.app
 
+# Backups
+*.bak
+
 # CMake files
 CMakeCache.txt
 CMakeFiles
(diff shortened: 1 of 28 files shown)


anton-malakhov:
replaced by #673

PR #562, opened 2020-06-08
WIP: Adding conda environment towards reproducible, isolated build

anton-malakhov:
Adding conda environment towards reproducible, isolated build
Derived from conda-forge/omniscidb-cpu-feedstock but moving
 all the workarounds into CMakeList.txt.
Added target that help preventing races before running main build.
conda recipe in OmniSciDB repo serves for easy development and can serve as an example for maintainers of feedstock recipes, not for publishing its builds directly.


scripts/conda/build.sh
```diff
@@ -0,0 +1,83 @@
...
+        export CXX=$HOST-g++
+        export CMAKE_CC=$HOST-gcc
+        export CMAKE_CXX=$HOST-g++
+    fi
+
+    GXX=$HOST-g++         # replace with $GXX
+    GCCVERSION=$(basename $(dirname $($GXX -print-libgcc-file-name)))
+fi
+
+set CMAKE_COMPILERS="-DCMAKE_C_COMPILER=$CMAKE_CC -DCMAKE_CXX_COMPILER=$CMAKE_CXX"
```
anton-malakhov, line 52: delete, unused

scripts/conda/build.sh
```diff
@@ -0,0 +1,83 @@
...
+     make --trace  # running sequentualy and enabling trace in case of failures
+make install || exit 1
+
+# copy initdb to mapd_initdb to avoid conflict with psql initdb
+mv $PREFIX/bin/initdb $PREFIX/bin/omnisci_initdb
+cd ..
+mkdir data
+# do lightweight testing here, make sanity_tests should go to under test env
+omnisci_initdb -f data
+omnisci_server --db-query-list SampleData/db-query-list-flights.sql --exit-after-warmup
```
anton-malakhov, line 81: unfortunately, doesn't work yet, with problem with thrift: ```
2020-06-08T17:22:01.744495 E 14055 DBHandler.cpp:5236 Exception: Exception occurred: org.apache.calcite.runtime.CalciteContextException: From line 2, column 6 to line 2, column 14: Object 'flights' not found
2020-06-08 17:22:01 ERROR TThreadPoolServer:run:315 - Thrift error occurred during processing of message.
org.apache.thrift.transport.TTransportException
        at org.apache.thrift.transport.TIOStreamTransport.read(TIOStreamTransport.java:132)
        at org.apache.thrift.transport.TTransport.readAll(TTransport.java:86)
        at org.apache.thrift.protocol.TBinaryProtocol.readAll(TBinaryProtocol.java:425)
        at org.apache.thrift.protocol.TBinaryProtocol.readI32(TBinaryProtocol.java:321)
        at org.apache.thrift.protocol.TBinaryProtocol.readMessageBegin(TBinaryProtocol.java:225)
        at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:27)
        at org.apache.thrift.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:310)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

**Comment by pearu** [comment]
You should really use https://github.com/conda-forge/omniscidb-feedstock as a template instead of omniscdb-cpu-feedstock.

**Comment by pearu** [comment]
> conda recipe in OmniSciDB repo serves for easy development and can serve as an example for maintainers of feedstock recipes

The conda-forge omniscidb-feedstock is ahead of this goal, imho.

**Comment by anton-malakhov** [comment]
> > conda recipe in OmniSciDB repo serves for easy development and can serve as an example for maintainers of feedstock recipes
> 
> The conda-forge omniscidb-feedstock is ahead of this goal, imho.
wow.. I was not aware of this feedstock. Thanks!
I'd not say that my PR is completely covered by this feedstock though. I suggest adding conda env file which is more convenient for development and debugging. Also there are certain changes to cmake to prevent races in parallel build and fix various issues right in CMakeLists, which I believe is a better fit to managing compiler settings. if you are interested in some of these improvements, please list them here and I'll rework this PR accordingly

**Comment by pearu** [comment]
> I suggest adding conda env file which is more convenient for development and debugging.

See also
https://github.com/Quansight/pearu-sandbox/blob/master/conda-envs/omniscidb-dev.yaml
https://github.com/Quansight/pearu-sandbox/blob/master/working-envs/activate-omniscidb-internal-dev.sh
that I am using for developing omniscidb within a conda environment.

It would be nice indeed if omniscidb could be developed without needing to change the conda environment as done in the `activate-omniscidb-internal-dev.sh` script.

**Comment by pearu** [comment]
> if you are interested in some of these improvements, ...

Yes! The ultimate goal would be that the build script of omniscidb-feedstock would need to have only the basic development steps:
```
mkdir build && cd build
cmake -DENABLE_CUDA=<on or off> ..
make -j $CPU_COUNT
```
where the cmake would recognize that the build happens from the conda environment and it would adjust the cmake variables according to how it is done in omniscidb-feedstock build script currently.
Thanks for your efforts!

File: scripts/conda/build.sh (hunk 2 above)
Comment (anton-malakhov) [reply scripts/conda/build.sh:82]: some issues fixed but still doesn't work, I'll update it later

**Comment by alexbaden** [comment]
@andrewseidl @pearu @anton-malakhov are we happy with these changes?

**Comment by anton-malakhov** [comment]
> @andrewseidl @pearu @anton-malakhov are we happy with these changes?

Not yet, we received those comments above and still want to converge with what @pearu is doing.
Also, we have another engineer, who starts looking into this task. I'll stay here keeping an eye on what he's doing.

**Comment by anton-malakhov** [comment]
Hi @pearu, @alexbaden  so as you know, I'm working on embedded database based on Omnisci. I need to pack yet another binaries in addition to the main omnisci package. I don't want to duplicate files though, but rather introduce [multiple output](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html?highlight=outputs#outputs-section) to the omnisci build recipe. What do you say if we move common stuff from opt/*  into share/omniscidb/ and create the following packages as sketched here:
```
outputs:
  - name omniscidb-common
    files:
      - share/omniscidb/calcite-1.0-SNAPSHOT-jar-with-dependencies.jar
      - share/omniscidb/QueryEngine/*
      - anything-else?
  - name: omniscidb-cpu
     files:
      - share/omniscidb/cpu/
  - name: omniscidb-gpu
     files:
      - share/omniscidb/gpu/
```
Also, my work will be about the following packages:
```
  - name: omniscidbe
    files:
      - lib/libDBEngine.so
      - include/DBEngine.*
    requirements:
      run:
        - {{ pin_subpackage('omniscidbe', exact=True) }}

  - name: omniscidbe4py
```
What do you say? Thanks

File: CMakeLists.txt [hunk 3]
```diff
@@ -1,6 +1,24 @@
...
+  add_definitions("-DBOOST_ERROR_CODE_HEADER_ONLY")
+  # Adding formating macros
+  add_definitions("-D__STDC_FORMAT_MACROS=1")
+  # fixes always_inline attribute errors
+  add_compile_options("-fno-semantic-interposition")
+  # Adding `--sysroot=...` resolves `no member named 'signbit' in the global namespace` error:
+  set(CMAKE_SYSROOT "$ENV{CONDA_BUILD_SYSROOT}")
+  set(CMAKE_PREFIX_PATH "$ENV{CONDA_PREFIX}")
+  # Adding `-I$BUILD_SYSROOT_INLCUDE` resolves `cassert file not found` error:
+  include_directories(SYSTEM "$ENV{CONDA_BUILD_SYSROOT}/usr/include")
```
pearu, line 19: The comment adds `BUILD_SYSROOT_INLCUDE` but actually adding `CONDA_BUILD_SYSROOT/usr/include`. Fix comment?

CMakeLists.txt
```diff
@@ -130,13 +148,14 @@ if(XCODE)
 endif()
 
 # fixme: hack works for Homebrew, might not work for Conda
-if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND NOT ENABLE_CONDA)
```
pearu, line 151: When conda enabled, use:
```
set(OPENSSL_ROOT_DIR "$ENV{CONDA_PREFIX}")
```
(see https://github.com/Quansight/omniscidb-cpu-feedstock/blob/master/recipe/build.sh#L46)

CMakeLists.txt
pearu: Considering https://github.com/conda-forge/omniscidb-feedstock/blob/master/recipe/build.sh, I am not sure if this is needed.

CMakeLists.txt
```diff
@@ -802,7 +814,7 @@ install(FILES omnisci.thrift DESTINATION ".")
 install(FILES common.thrift DESTINATION ".")
 install(FILES QueryEngine/serialized_result_set.thrift DESTINATION "QueryEngine/")
 
-if(NOT PREFER_STATIC_LIBS)
+if(NOT PREFER_STATIC_LIBS AND NOT ENABLE_CONDA)
```
pearu, line 817: Thumbs up!
Btw, is `install(FILES ${Boost_LIBRARIES} DESTINATION ThirdParty/lib)` really needed?

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# go overwrites CC and CXX with nonsense (see
+# https://github.com/conda-forge/go-feedstock/issues/47), hence we
+# redefine these below. Reset GO env variables for omniscidb build
+# (IIRC, it is needed for CUDA support):
+export CGO_ENABLED=1
+export CGO_LDFLAGS=
+export CGO_CFLAGS=$CFLAGS
+export CGO_CPPFLAGS=
+
```
pearu, line 18: https://github.com/conda-forge/go-feedstock/issues/47 is resolved, so we should be able to delete the above.

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+    # clang.
+    COMPILERNAME=gcc                      # options: clang, gcc
+
+    if [ "$COMPILERNAME" == "clang" ]; then
+        # All these must be picked up from $PREFIX/bin
+        export CC=clang
+        export CXX=clang++
+        export CMAKE_CC=clang
+        export CMAKE_CXX=clang++
+    else
```
pearu, line 41: I suggest not to use clang as a compiler when using conda environment under Linux. Hence, delete the above.

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+        export CC=$HOST-gcc
+        export CXX=$HOST-g++
+        export CMAKE_CC=$HOST-gcc
+        export CMAKE_CXX=$HOST-g++
+    fi
```
pearu, line 46: Not needed as CC and CXX should already point to correct conda compilers.

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+    GXX=$HOST-g++         # replace with $GXX
+    GCCVERSION=$(basename $(dirname $($GXX -print-libgcc-file-name)))
+fi
+
+set CMAKE_COMPILERS="-DCMAKE_C_COMPILER=$CMAKE_CC -DCMAKE_CXX_COMPILER=$CMAKE_CXX"
+
+cmake -S . -B build -Wno-dev \
+    -DCMAKE_INSTALL_PREFIX="$PREFIX"\
+    -DCMAKE_C_COMPILER=$CMAKE_CC\
+    -DCMAKE_CXX_COMPILER=$CMAKE_CXX\
```
pearu, line 57: Use CC and CXX as in https://github.com/conda-forge/omniscidb-feedstock/blob/master/recipe/build.sh#L27

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+    fi
+
+    GXX=$HOST-g++         # replace with $GXX
+    GCCVERSION=$(basename $(dirname $($GXX -print-libgcc-file-name)))
+fi
+
+set CMAKE_COMPILERS="-DCMAKE_C_COMPILER=$CMAKE_CC -DCMAKE_CXX_COMPILER=$CMAKE_CXX"
+
+cmake -S . -B build -Wno-dev \
+    -DCMAKE_INSTALL_PREFIX="$PREFIX"\
```
pearu, line 55: Use some INSTALL_BASE (opt or shared) as in https://github.com/conda-forge/omniscidb-feedstock/blob/master/recipe/build.sh#L97

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+    -DCMAKE_CXX_COMPILER=$CMAKE_CXX\
+    -DCMAKE_BUILD_TYPE=release\
+    -DMAPD_DOCS_DOWNLOAD=OFF\
+    -DENABLE_AWS_S3=OFF\
+    -DENABLE_CUDA=OFF\
+    -DENABLE_FOLLY=OFF\
+    -DENABLE_JAVA_REMOTE_DEBUG=OFF\
+    -DENABLE_PROFILER=OFF\
+    -DENABLE_TESTS=OFF\
+    -DPREFER_STATIC_LIBS=OFF\
```
pearu, line 66: For CUDA builds, add also `-DENABLE_CUDA=on-DCUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME ` as in https://github.com/conda-forge/omniscidb-feedstock/blob/master/recipe/build.sh#L54

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+    || exit 1
+
+pushd build
+# preventing races by executing vulnerable targets first
+make PatchParser PatchScanner thrift_gen -j && make -j || \
+     make --trace  # running sequentualy and enabling trace in case of failures
```
pearu, line 72: This should not be needed anymore.

scripts/conda/build.sh
```diff
@@ -0,0 +1,84 @@
...
+make install || exit 1
+
+# copy initdb to mapd_initdb to avoid conflict with psql initdb
```
pearu, line 75: replace mapd->omnisci

scripts/conda/meta.yaml
```diff
@@ -0,0 +1,77 @@
+{% set name = "omniscidb-cpu" %}
```
pearu, line 1: I suggest copying the content from https://github.com/conda-forge/omniscidb-feedstock/blob/master/recipe/meta.yaml rather than from the old omniscidb-cpu-feedstock.

scripts/mapd-deps-conda-dev-env.yml
```diff
@@ -0,0 +1,40 @@
+#  conda env create -f scripts/mapd-deps-conda-dev-env.yml --force
```
pearu, line 1: Replace this (out-dated) file with
https://github.com/Quansight/pearu-sandbox/blob/master/conda-envs/omniscidb-dev.yaml

anton-malakhov:
thank you, @pearu, I'm reworking the PR right now! Do you have some considerations w.r.t. multiple packages and installation path as I wrote before? Thanks!

pearu:
> Do you have some considerations w.r.t. multiple packages and installation path as I wrote before?

Using outputs is fine. However, in omniscidb-feedstock we use name `omniscidb` with cpu/cuda build string rather than `omniscidb-cpu` and `omniscidb-gpu` to distinguish cuda disabled and enabled builds. Could you do the same?

Re using `share` instead of `opt`: Note that "The /usr/share hierarchy is for all read-only architecture independent data files", hence `share` cannot contain omniscidb binaries, imho.
Copying `.jar` and `.thrift` files to `share` sounds good to me.

Btw, `gpu != cuda`, so I suggest avoid using `gpu` to represent the `cuda` target.

anton-malakhov:
@pearu here is a draft of what I want to do in this PR as well: https://github.com/intel-go/omniscidb/pull/73:
i.e. mark cmake components and split this package using conda's multiple output. The 'dbe' and 'dbe4py' parts are our internal stuff yet. But they can illustrate why I want do this sort of things. Do you see any design showstoppers on this way?

CMakeLists.txt
```diff
@@ -1,6 +1,24 @@
...
+set(ENABLE_CONDA OFF)
+if(DEFINED ENV{CONDA_PREFIX})
+  set(ENABLE_CONDA ON)
+  # various fixes and workarounds
+  add_definitions("-Dsecure_getenv=getenv")
+  # fixes `undefined reference to `boost::system::detail::system_category_instance'`:
+  add_definitions("-DBOOST_ERROR_CODE_HEADER_ONLY")
+  # Adding formating macros
+  add_definitions("-D__STDC_FORMAT_MACROS=1")
+  # fixes always_inline attribute errors
```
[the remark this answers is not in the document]
anton-malakhov: Used to be a problem, but I don't see it anymore

anton-malakhov:
It is still WIP which doesn't even work yet. Though, some help is welcome

anton-malakhov:
> Using outputs is fine. However, in omniscidb-feedstock we use name omniscidb with cpu/cuda build string rather than omniscidb-cpu and omniscidb-gpu to distinguish cuda disabled and enabled builds. Could you do the same?

I'm not sure if I can do that in multiple outputs mode. My idea is to separate shared content (-common) from cpu, dbe and gpu packages. May be gpu should be a generic version.. while cpu goes as cpu-only version. If building gpu version requires separate run of conda-build, it looks suboptimal because -common part will be produced twice.
Can you do multiple outputs in conda-forge version as well? and how you'd do that?

pearu:
> I'm not sure if I can do that in multiple outputs mode...

The shared output (-common) can be generated only in one conda build, in other builds the generating shared output will be skipped. See https://github.com/conda-forge/arrow-cpp-feedstock/blob/master/recipe/meta.yaml as an example of conda-forge multiple builds, notice the usage of `skip` in `build` sections. HTH.

CMakeLists.txt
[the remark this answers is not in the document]
anton-malakhov: let's keep it at least because it can improve performance of shared libraries as explained here: https://stackoverflow.com/a/48521068/2527797

PR #585, opened 2020-08-26
Introduce COMPONENT labels, improving conda build support

anton-malakhov:
This is cmake-only continuation of #562, which fixes cmake/conda build problems.
Labels are needed for convenient splitting of the installation files into different packages.

@pearu I hope I have not missed your comments in this PR.


anton-malakhov:
a kind ping, @alexbaden @pearu

PR #599, opened 2020-10-30
FSI fixes and improvements

fexolm:
I'm sorry that we haven't split that changes to separate commits, but it was a bit difficult after we merge the master branch. 
So, the major changes we did here are:
- The main change we have done is splitting Arrow FSI to ArrowCsvForeignStorage, ArrowForeignStorage, the first one is for `CREATE DATAFRAME` query, and the second one is for our embedded engine, to allow users passing arrow tables directly to omnisci engine. 
- Refactor Arrow FSI a bit, so fragmenting logic and updating metadata is done only in one place now.  
- Refactor threading a bit, to get rid of task groups, so we have a parallel for a loop by columns.
- Add initial implementation for converting arrow's dictionaries to omnisci dictionaries.


fexolm:
Formatting issues seem to be out of scope of that pr.

PR #601, opened 2020-11-09
Modify calcite server to execute calcite ast queries

fexolm:


java/calcite/src/main/java/com/mapd/parser/server/CalciteServerHandler.java
```diff
@@ -175,10 +182,29 @@ public TPlanResult process(String user,
...
+          // also resolve all the views in the select part
+          // resolution of the other parts is not
+          // necessary as these cannot be views
+          resolvedAccessedObjects.tables_selected_from = new ArrayList<>(parser.resolveSelectIdentifiers(capturer));
+          resolvedAccessedObjects.tables_inserted_into = new ArrayList<>(capturer.inserts);
+          resolvedAccessedObjects.tables_updated_in = new ArrayList<>(capturer.updates);
+          resolvedAccessedObjects.tables_deleted_from = new ArrayList<>(capturer.deletes);
+
+        } else {
+          jsonResult = parser.optimizeRAQuery(queryText);
```
yoonminnam, line 206: 1) Why RAQuery does not need to specify `primaryAccessedObjects` and `resolvedAccessedObjects`?

2) Also, if RAQuery only needs to apply some specific optimization rules, then can we completely separate the logic to keep the current OmniSci Calcite server logic as stable as possible to make the least change?

Like..

if (isRAQuery) {
    jsonResult = parser.optimizeRAQuery(queryText);
} else {
  // put the existing OmniSci parsing logic --
try {
       ...
       try {
         ...
       } catch(...) { ... }
  } catch(...) { ... }
}

java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
```diff
@@ -458,7 +465,33 @@ RexNode toRex(RelInput relInput, Object o) {
         if (literal == null) {
           return rexBuilder.makeNullLiteral(sqlTypeName);
         }
-        return toRex(relInput, literal);
+
+        // omnisci serializes numeric literals differently, we need more data
```
yoonminnam, line 469: Do we confirm that this addition has any negative issue on the existing OmniSci engine?

yoonminnam (changes requested):
Overall, this change makes sense to me up to a certain point.
But I think we have to check that the changes and the addition to the existing Calcite logic does not bring the negative impact on the OmniSci-side operation.

java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
yoonminnam: For your convinience, I tested this branch through jenkins by cherry-picking all changes into the latest master. 
And it passes all tests.

java/calcite/src/main/java/com/mapd/parser/server/CalciteServerHandler.java
fexolm: 1)I'm not sure about `primaryAccessedObjects` and `resolvedAccessedObjects`. I don't really understand what are they for. Also, we didn't found any case, where our query does not work without them. And it looks like they come from SQLNode, and I don't really see the easy way to extract them from calcite ast. 

2) I see, what you mean, will send a fix soon

fexolm: I've sent a commit with possible fix of 2

yoonminnam (approved):
LGTM.
But please check the formatting issue. The formatting test says (I don't know why but) this PR touches the file that is not related to this PR.

fexolm:
It looks like, current master has inconsistent formatting, maybe it is better to fix formatting in the other pr.

java/calcite/src/main/java/org/apache/calcite/prepare/MapDPlanner.java
```diff
@@ -192,6 +210,42 @@ private RelRoot applyQueryOptimizationRules(RelRoot root) {
...
+    RelNode newRel = relR.project();
+
+    do {
+      oldRel = newRel;
+      newRel = program.run(null,
+              oldRel,
+              null,
+              ImmutableList.<RelOptMaterialization>of(),
+              ImmutableList.<RelOptLattice>of());
+      // there must be a better way to compare these
```
alexbaden, line 243: This looks like legacy code -- we have since modified some of these optimizations to be a bit cleaner and use built in Calcite rules, see https://github.com/omnisci/omniscidb/blob/master/java/calcite/src/main/java/com/mapd/calcite/parser/MapDParser.java#L821

java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
```diff
@@ -458,7 +465,33 @@ RexNode toRex(RelInput relInput, Object o) {
...
+          final long scale = ((Number) map.get("scale")).longValue();
+          final long precision = ((Number) map.get("precision")).longValue();
+          final long typeScale = ((Number) map.get("type_scale")).longValue();
+          final long typePrecision = ((Number) map.get("type_precision")).longValue();
+          RelDataTypeFactory typeFactory = cluster.getTypeFactory();
+
+          BigDecimal value =
+                  BigDecimal.valueOf(((Number) literal).longValue(), (int) scale);
+
+          if (typeScale != 0 && typeScale != -2147483648) {
```
alexbaden, line 483: Surprised to see the null sentinel here in the `type_scale` field. Can you explain?

alexbaden (approved):
Two comments but nothing blocking, we will get this merged today.

java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
fexolm: As we found out, omnisci format has type_scale field even if literal is not decimal. At that case typeScale field would have the value of -2147483648.
So, we have the same logic in our calcite serializer [here](https://github.com/modin-project/modin/blob/623de5c3c90f9d44262f5ce4e9d7bae6d54e5dc8/modin/experimental/engines/omnisci_on_ray/frame/calcite_serializer.py#L132).

java/calcite/src/main/java/org/apache/calcite/prepare/MapDPlanner.java
fexolm: I see, thanks. I think, I would better send the fix as separate pr, as we need to spend some time testing new logic.

PR #615, opened 2021-01-15
Enable ITT notify profiling API for queries and keywords

anton-malakhov:
VTune-enabled build. It is now possible to control the VTune traces via `--itt_pause` & `--itt_resume` in warmup queries


vnlitvinov:
Note that ITT API is available as conda package here: https://anaconda.org/conda-forge/ittapi, and as source here: https://github.com/intel/ittapi

PR #623, opened 2021-02-16
Add option to set Monday as first day of week

abykovsk:
Added option to set Monday as first day of week


mattpulver:
Thank you for the PR.

Few thoughts on this:
* In general we prefer to stay away from configuration options that produce different results. This makes it harder to debug query results from customers that may have different settings.
* There are other contexts of what it means to be the beginning of a week. For example `EXTRACT(WEEK ...)` has a separate meaning for what the first day of the week is, which remains unaffected, that this setting needs to be differentiated from in the `--help` message.
* This is replicated with `EXTRACT(ISODOW FROM timestamp) - 1`. Would this suffice?

simoneves:
Closing as answered with no response

PR #624, opened 2021-02-16
Add some datetime extraction functions

abykovsk:
Added functions to extract last day of month, determine leap years, determine end/start of month/quarter/year.


mattpulver:
Thank you for the PR.

Are these features being required/requested in a real project? If so a brief description to motivate these features would be useful. In general we don't want to add features that may look nice but are not used in practice.
￼

mflaxman10:
Can vouch for at least two customer requests for the quarter-related functions, which are of course common in business reporting.

PR #625, opened 2021-03-01
WIP: level zero & spv translator build integration

kurapov-peter:
Hi,

I was able to modify mapd-deps for ubuntu from-source build to include spirv translator in-tree llvm-10 build (static linkage only). There're 2 simple tests that show components work properly that pass for me. Notice that I had to link the spv test with thrift handler for some reason - otherwise getting linking issue:
```
/home/petr/code/offloading/omniscidb/ThriftHandler/QueryState.cpp:51: error: undefined reference to 'Catalog_Namespace::UserMetadata::userLoggable[abi:cxx11]() const'
/home/petr/code/offloading/omniscidb/ThriftHandler/QueryState.cpp:164: error: undefined reference to 'Catalog_Namespace::UserMetadata::userLoggable[abi:cxx11]() const'
```

FindLevelZero is taken from here (there's no single source for that for now, so just copying): https://github.com/oneapi-src/level-zero-tests/blob/master/cmake/FindLevelZero.cmake


kurapov-peter:
Level zero build is under an option:
cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_L0=ON -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=gold -DENABLE_AWS_S3=off
Works along with cuda.

kurapov-peter:
These changes will go along with others in another pr

PR #628, opened 2021-03-10
Use tbb::parallel_sort instead of std::sort.

ienkovich:
Trivial change to get parallel sort. Get 4x speed-up for TCP-H Q11 on scale=500.

### DIFF
--- a/QueryEngine/ResultSet.cpp
+++ b/QueryEngine/ResultSet.cpp
@@ -37,6 +37,10 @@
 #include "Shared/thread_count.h"
 #include "Shared/threadpool.h"
 
+#ifdef HAVE_TBB
+#include "tbb/parallel_sort.h"
+#endif
+
 #include <algorithm>
 #include <bitset>
 #include <future>
@@ -546,7 +550,8 @@ void ResultSet::sort(const std::list<Analyzer::OrderEntry>& order_entries,
     if (top_n == 0) {
       top_n = pv.size();  // top_n == 0 implies a full sort
     }
-    pv = topPermutation(pv, top_n, createComparator(order_entries, pv, executor, false));
+    pv = topPermutation(
+        pv, top_n, createComparator(order_entries, pv, executor, false), false);
     if (pv.size() < permutation_.size()) {
       permutation_.resize(pv.size());
       permutation_.shrink_to_fit();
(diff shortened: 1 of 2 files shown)


yoonminnam:
@ienkovich can you rebase this?

ienkovich:
> @ienkovich can you rebase this?

Done

mattpulver:
`ResultSet::topPermutation` is called in a few places, one of which is already done from a thread inside of a multi-threaded context in `ResultSet::parallelTop()`:
```
  const auto top_sort_interval = [&, top_n, executor](const auto interval) {
    PermutationView pv(permutation_.data() + interval.begin, 0, interval.size());
    pv = initPermutationBuffer(pv, interval.begin, interval.end);
    const auto compare = createComparator(order_entries, pv, executor, true);
    permutation_views[interval.index] = topPermutation(pv, top_n, compare);
  };
  threadpool::FuturesThreadPool<void> top_sort_threads;
  for (auto interval : makeIntervals<PermutationIdx>(0, permutation_.size(), nthreads)) {
    top_sort_threads.spawn(top_sort_interval, interval);
  }
  top_sort_threads.join();
```
Will calling `tbb::parallel_sort()` from within this context help or hinder performance? The number of threads being used in this case is already equal to `nthreads = cpu_threads()`.

mattpulver:
The call to `ResultSet::topPermutation()` from https://github.com/intel-ai/omniscidb/blob/ienkovich/tbb-parallel_sort/QueryEngine/ResultSet.cpp#L613 is already done in a multi-threaded context using all available cpu threads. Will calling `tbb:parallel_sort()` help or hurt performance in this case?

ienkovich:
> The call to `ResultSet::topPermutation()` from https://github.com/intel-ai/omniscidb/blob/ienkovich/tbb-parallel_sort/QueryEngine/ResultSet.cpp#L613 is already done in a multi-threaded context using all available cpu threads. Will calling `tbb:parallel_sort()` help or hurt performance in this case?

I see there were significant changes since I've created the original patch. TBB has a composability feature that allows nested parallelism levels with no harm, but here we use `FuturesThreadPool` unconditionally (even if TBB is enabled) and this might result in some performance issues. Please give me some time to rethink and retest this patch.

ienkovich:
@mattpulver I added a flag to `ResultSet::topPermutation()` to allow multiple threads usage when we don't use it within a thread pool.

PR #638, opened 2021-03-26
Fix DBE build

igalink:
Fix DBE build: upadte DBHandler init parameters and return removed ExecuteDDL function.
    Fix segfault on std::filesystem::path destructor by changing gcc version from 8 to 9.


alexbaden:
It is possible GCC 9 will break our CUDA builds, so we're doing a bit of investigation there -- otherwise this looks fine.

igalink:
This PR updates GCC for conda environment only. It doesn't affect standard OmnicsiDB build.

PR #644, opened 2021-04-05
Shared execution context draft

ienkovich:
A shared execution context is used to reduce initialization and reduction costs. The feature is in a draft state, works only for a keyless perfect hash with COUNT and SUM aggregates. Works with sub-fragments only.

### DIFF
--- a/QueryEngine/Descriptors/QueryMemoryDescriptor.cpp
+++ b/QueryEngine/Descriptors/QueryMemoryDescriptor.cpp
@@ -27,6 +27,7 @@
 
 bool g_enable_smem_group_by{true};
 extern bool g_enable_columnar_output;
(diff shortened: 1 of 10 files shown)


tmostak:
@ienkovich how is this branch looking in terms of moving out of draft? Also saw the asan failure in GeospatialTest, any ideas on that?

ienkovich:
> @ienkovich how is this branch looking in terms of moving out of draft? Also saw the asan failure in GeospatialTest, any ideas on that?

The functionality of this patch is quite limited but it passes tests and can be used for experiments and further extension.

Asan failure description is not very helpful. Right now can't see anything in the patch which could cause such an error, especially with the feature disabled during testing.

PR #646, opened 2021-04-15
Set root_abs_path relative to libDBEngine

igalink:
Set root_abs_path relative to libDBEngine in the case of a call from the built-in library


pearu:
Superseded by #647 .

PR #647, opened 2021-04-18
Support omniscidb conda-forge feedstock builds with embedding feature [Linux-only]

pearu:
This PR includes:
- determine the omniscidb root path from the location of libDBEngine.so
- include a complete set of thrift files to installation
- use OMNISCI_ROOT_PATH environment variable to determine the root path for building Python omniscidbe extension module
- Python extension module `dbe` is renamed to `omniscidbe`

This PR is used in https://github.com/conda-forge/omniscidb-feedstock/pull/39 to build the following conda-forge packages:
- omniscdb-common, linux only
- omniscidb, cpu and cuda builds, linux only
- omniscidbe, cpu and cuda builds, linux only
- pyomniscidb, python versions 3.6, 3.7, 3.8, linux, windows, os
- pyomniscidbe, cpu and cuda builds, python version 3.6, 3.7, 3.8, linux only

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,6 +338,7 @@ if(ENABLE_DBE)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
   set(ENABLE_STANDALONE_CALCITE ON)
   add_definitions("-DENABLE_EMBEDDED_DATABASE")
+  add_definitions("-DDBEngine_LIBNAME=\"${CMAKE_SHARED_LIBRARY_PREFIX}DBEngine${CMAKE_SHARED_LIBRARY_SUFFIX}\"")
 endif()
 
 # Code coverage
@@ -854,6 +855,8 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/ThirdParty/licenses" DESTINATION 
 
 # OmniSciTypes.h local includes (for UDF/UDTF)
 install(FILES Shared/funcannotations.h DESTINATION "Shared/" COMPONENT "include")
+install(FILES Shared/InlineNullValues.h DESTINATION "Shared/" COMPONENT "include")
+install(FILES Logger/Logger.h DESTINATION "Logger/" COMPONENT "include")
 
 # Frontend
 option(MAPD_IMMERSE_DOWNLOAD "Download OmniSci Immerse for packaging" OFF)
@@ -969,6 +972,7 @@ install(FILES completion_hints.thrift DESTINATION "." COMPONENT "thrift")
 install(FILES omnisci.thrift DESTINATION "." COMPONENT "thrift")
 install(FILES common.thrift DESTINATION "." COMPONENT "thrift")
 install(FILES QueryEngine/serialized_result_set.thrift DESTINATION "QueryEngine/" COMPONENT "thrift")
+install(FILES QueryEngine/extension_functions.thrift DESTINATION "QueryEngine/" COMPONENT "thrift")
 
 if(NOT PREFER_STATIC_LIBS AND NOT ENABLE_CONDA)
   install(FILES ${Boost_LIBRARIES} DESTINATION ThirdParty/lib)
--- a/Embedded/setup.in.py
+++ b/Embedded/setup.in.py
@@ -16,10 +16,10 @@ root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # conda-forge packages omniscidbe and pyomniscidbe are built
 # separately. OMNISCI_ROOT_PATH is defined by the omniscidbe activate
 # script that determines the location of libDBEngine.so and is
-# required in linking Python extension module omniscidbe.
-omnisci_root_path = os.environ.get('OMNISCI_ROOT_PATH',
-                                   os.path.join(sys.exec_prefix, 'opt', 'omnisci'))
-omniscidbe_root = os.path.join(omnisci_root_path, 'lib')
+# required for linking Python extension module omniscidbe.
+extra_library_dirs = []
+if 'OMNISCI_ROOT_PATH' in os.environ:
+    extra_library_dirs.append(os.path.join(os.environ['OMNISCI_ROOT_PATH'], 'lib'))
 
 dbe = Extension(
     "omniscidbe",
@@ -34,8 +34,8 @@ dbe = Extension(
         "@CMAKE_SOURCE_DIR@/ThirdParty/rapidjson",
         "@CMAKE_SOURCE_DIR@/Distributed/os",
     ],
-    library_dirs=pa.get_library_dirs() + ["@CMAKE_CURRENT_BINARY_DIR@", ".", omniscidbe_root],
-    runtime_library_dirs=pa.get_library_dirs() + ["$ORIGIN/../../", omniscidbe_root],
+    library_dirs=pa.get_library_dirs() + ["@CMAKE_CURRENT_BINARY_DIR@", "."] + extra_library_dirs,
+    runtime_library_dirs=pa.get_library_dirs() + ["$ORIGIN/../../"] + extra_library_dirs,
     libraries=pa.get_libraries() + ["DBEngine", "boost_system"],
     extra_compile_args=["-std=c++17", "-DRAPIDJSON_HAS_STDSTRING"],
 )
--- a/Embedded/setup.in.py
+++ b/Embedded/setup.in.py
@@ -7,12 +7,22 @@ from Cython.Build import cythonize
 from distutils.core import setup, Extension
 
 import os
+import sys
 import numpy as np
 import pyarrow as pa
 
 root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+# conda-forge packages omniscidbe and pyomniscidbe are built
+# separately. OMNISCI_ROOT_PATH is defined by the omniscidbe activate
+# script that determines the location of libDBEngine.so and is
+# required in linking Python extension module omniscidbe.
+omnisci_root_path = os.environ.get('OMNISCI_ROOT_PATH',
+                                   os.path.join(sys.exec_prefix, 'opt', 'omnisci'))
+omniscidbe_root = os.path.join(omnisci_root_path, 'lib')
+
 dbe = Extension(
-    "dbe",
+    "omniscidbe",
     ["@CMAKE_CURRENT_SOURCE_DIR@/Python/dbe.pyx"],
     language="c++17",
     include_dirs=[
@@ -24,8 +34,8 @@ dbe = Extension(
         "@CMAKE_SOURCE_DIR@/ThirdParty/rapidjson",
         "@CMAKE_SOURCE_DIR@/Distributed/os",
     ],
-    library_dirs=pa.get_library_dirs() + ["@CMAKE_CURRENT_BINARY_DIR@", "."],
-    runtime_library_dirs=pa.get_library_dirs() + ["$ORIGIN/../../"],
+    library_dirs=pa.get_library_dirs() + ["@CMAKE_CURRENT_BINARY_DIR@", ".", omniscidbe_root],
+    runtime_library_dirs=pa.get_library_dirs() + ["$ORIGIN/../../", omniscidbe_root],
     libraries=pa.get_libraries() + ["DBEngine", "boost_system"],
     extra_compile_args=["-std=c++17", "-DRAPIDJSON_HAS_STDSTRING"],
 )
@@ -59,7 +69,7 @@ if False:  # TODO: implement an option?
     ]
 
 setup(
-    name="dbe",
+    name="omniscidbe",
     version="0.1",
     ext_modules=cythonize(
         dbe,
--- a/OSDependent/Unix/omnisci_path.cpp
+++ b/OSDependent/Unix/omnisci_path.cpp
@@ -26,10 +26,30 @@
 #include <boost/filesystem/path.hpp>
 
 #include "Logger/Logger.h"
+#ifdef ENABLE_EMBEDDED_DATABASE
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <dlfcn.h>
+#include <link.h>
+#endif
 
 namespace omnisci {
 
 std::string get_root_abs_path() {
+#ifdef ENABLE_EMBEDDED_DATABASE
+  void* const handle = dlopen(DBEngine_LIBNAME, RTLD_LAZY);
+  CHECK(handle);
+  const struct link_map* link_map = 0;
+  const int ret = dlinfo(handle, RTLD_DI_LINKMAP, &link_map);
+  CHECK_EQ(ret, 0);
+  CHECK(link_map);
+  /* Despite the dlinfo man page claim that l_name is absolute path,
+     it is so only when the location path to the library is absolute,
+     say, as specified in LD_LIBRARY_PATH. */
+  boost::filesystem::path abs_exe_dir(boost::filesystem::absolute(
+      boost::filesystem::canonical(std::string(link_map->l_name))));
+#else
 #ifdef __APPLE__
   char abs_exe_path[PROC_PIDPATHINFO_MAXSIZE] = {0};
   auto path_len = proc_pidpath(getpid(), abs_exe_path, sizeof(abs_exe_path));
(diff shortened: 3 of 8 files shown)


pearu:
@igalink @Garra1980 @vnlitvinov, please review.

pearu:
@igalink @Garra1980 @vnlitvinov, please review.

Embedded/setup.in.py
```diff
@@ -24,8 +34,8 @@
         "@CMAKE_SOURCE_DIR@/ThirdParty/rapidjson",
         "@CMAKE_SOURCE_DIR@/Distributed/os",
     ],
-    library_dirs=pa.get_library_dirs() + ["@CMAKE_CURRENT_BINARY_DIR@", "."],
-    runtime_library_dirs=pa.get_library_dirs() + ["$ORIGIN/../../"],
+    library_dirs=pa.get_library_dirs() + ["@CMAKE_CURRENT_BINARY_DIR@", ".", omniscidbe_root],
+    runtime_library_dirs=pa.get_library_dirs() + ["$ORIGIN/../../", omniscidbe_root],
```
vnlitvinov, line 38: why is this needed again?

vnlitvinov (approved):
LGTM, with one small question

Embedded/setup.in.py
pearu: This is required for building Python module omniscidbe from conda recipe. Notice that `libDBEngine.so` is provided by another conda package (`omniscidbe`) and is not available as `Embedded/`libDBEngine.so` that setup.py otherwise assumes.

There are actually two possible approaches to specify the path to `libDBEngine.so`:
1. `$CONDA_PREFIX/lib`
2. `$OMNISCI_ROOT_PATH/lib`

In this PR, option 2. is used as a more general one - it would be functional when using some non-conda environment.

CMakeLists.txt
```diff
@@ -854,6 +855,8 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/ThirdParty/licenses" DESTINATION
 
 # OmniSciTypes.h local includes (for UDF/UDTF)
 install(FILES Shared/funcannotations.h DESTINATION "Shared/" COMPONENT "include")
+install(FILES Shared/InlineNullValues.h DESTINATION "Shared/" COMPONENT "include")
```
Garra1980, line 858: Is it specific for DBEngine changes?

pearu: nope, these are the header files that OmniSciTypes.h include.

OSDependent/Unix/omnisci_path.cpp
```diff
@@ -26,10 +26,30 @@
...
+  const struct link_map* link_map = 0;
+  const int ret = dlinfo(handle, RTLD_DI_LINKMAP, &link_map);
+  CHECK_EQ(ret, 0);
+  CHECK(link_map);
+  /* Despite the dlinfo man page claim that l_name is absolute path,
+     it is so only when the location path to the library is absolute,
+     say, as specified in LD_LIBRARY_PATH. */
+  boost::filesystem::path abs_exe_dir(boost::filesystem::absolute(
+      boost::filesystem::canonical(std::string(link_map->l_name))));
+#else
```
Garra1980, line 52: So, no Windows support so far?

CMakeLists.txt
Garra1980: I'm just wondering should this go as standalone PR?

OSDependent/Unix/omnisci_path.cpp
pearu: AFAIK, not yet. In case I am mistaken, it is better to land this PR for Linux-only because creating conda packages of omniscidb for windows is a non-trivial task on its own.

Garra1980: Agree, can we reflect this somehow, eg in PR's name?

CMakeLists.txt
```diff
@@ -338,6 +338,7 @@ if(ENABLE_DBE)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
   set(ENABLE_STANDALONE_CALCITE ON)
   add_definitions("-DENABLE_EMBEDDED_DATABASE")
+  add_definitions("-DDBEngine_LIBNAME=\"${CMAKE_SHARED_LIBRARY_PREFIX}DBEngine${CMAKE_SHARED_LIBRARY_SUFFIX}\"")
```
Garra1980, line 341: What is this for ?Do we use DBEngine_LIBNAME somewhere?

CMakeLists.txt
pearu: Adding the missing header files as well as thrift files are simple bug fixes that are indirectly related to DBE packaging support (the omniscidb-feedstock would require these fixes anyway), and do not deserve a separate PR, IMHO.
If there is a particular need for the separation of these changesets then we can do it.

CMakeLists.txt
pearu: Yes, this is used in
https://github.com/omnisci/omniscidb/pull/647/files#diff-572dc5a7d5354cdeee74fa14ede04679a0fad4824c1153947f8d0e10e34e09e1R41

Garra1980: Ah, right, my bad

OSDependent/Unix/omnisci_path.cpp
pearu: Sure, done.

Btw, currently, all omniscidb conda packages, except pyomniscidb, are Linux-only.

Garra1980: Yeah, I think this is ok for now

CMakeLists.txt
Garra1980: No problem, I just worried a bit how omnisci is being built now without this changes :)

Garra1980:
I'm ok once python tests pass

Garra1980 (approved):

CMakeLists.txt
Garra1980: Great then, thanks for clarification

pearu:
> I'm ok once python tests pass

Locally, the three Python tests in `Embedded/test` pass ok.
These tests are now enabled also in https://github.com/conda-forge/omniscidb-feedstock/pull/39 but no confirmation yet.

CMakeLists.txt
igalink: I don’t think this is necessary because different OSDependent/Unix/omnisci_path.cpp files are used for different OS.

Embedded/setup.in.py
igalink: Could you please clarify the file location structure for the CF package build?

Embedded/setup.in.py
```diff
@@ -7,12 +7,22 @@
...
 import numpy as np
 import pyarrow as pa
 
 root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+# conda-forge packages omniscidbe and pyomniscidbe are built
+# separately. OMNISCI_ROOT_PATH is defined by the omniscidbe activate
+# script that determines the location of libDBEngine.so and is
+# required in linking Python extension module omniscidbe.
+omnisci_root_path = os.environ.get('OMNISCI_ROOT_PATH',
```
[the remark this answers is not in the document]
pearu: Sure, it is defined in omniscidbe activate script:

https://github.com/conda-forge/omniscidb-feedstock/pull/39/files#diff-950dd1a30450006d4e2daa139eb257c38e5ffa2cc5f0dabbe9b12dbf485f0d00R213

OSDependent/Unix/omnisci_path.cpp
```diff
@@ -26,10 +26,30 @@
...
+#endif
+#include <dlfcn.h>
+#include <link.h>
+#endif
 
 namespace omnisci {
 
 std::string get_root_abs_path() {
+#ifdef ENABLE_EMBEDDED_DATABASE
+  void* const handle = dlopen(DBEngine_LIBNAME, RTLD_LAZY);
```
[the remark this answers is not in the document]
pearu: To my understanding, `ENABLE_EMBEDDED_DATABASE` is effective only when building `libDBEngine.so`.

The omniscidb server itself does not load `libDBEngine.so`, hence the `dlopen` is never called in this case.

Let's consider an application that links against `libDBEngine.so` (such as the Python omniscidbe module). The `get_root_abs_path` function is defined in the `libDBEngine.so` library. When the application calls `get_root_abs_path`, it must mean that the `libDBEngine.so` library has been already loaded and the `dlopen` call here should be a no-op, IIUC.
So, the issue of loading unused library is not valid because calling `get_root_abs_path` is about using the library. Let me know if I misunderstood the issue.

If there is another way to acquire the handle of a loaded library, it could be used as well.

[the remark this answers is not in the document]
igalink: The get_root_abs_path function is defined in the source code which is used both for building a libDBEngine.so and a server, utilities (for example, initdb).
Actually, libDBEngine.so is built additionally to the standard omniscidb artifacts in case of ENABLE_EMBEDDED_DATABASE option.

Embedded/setup.in.py
pearu: Sure. All paths will be shown relative to conda environment `$CONDA_PREFIX`:

1. `omniscidb-common` defines the following paths
    ```
    opt/omnisci/bin/{calcite, omnisci}-*.jar
    opt/omnisci/QueryEngine/RuntimeFunctions.bc
    ...
    ```
2. `omniscidb` defines
    ```
    opt/omnisci/bin/{omnisci_server,omnisci_initdb,omnisql,...}
    ```
3. `omniscidbe` defines
    ```
    opt/omnisci/lib/libDBEngine.so
    lib/libDBEngine.so -> ../opt/omnisci/lib/libDBEngine.so
    ```
4. `pyomniscidbe` for Python 3.8 defines
    ```
    lib/python3.8/site-packages/omniscidbe.cpython-38-x86_64-linux-gnu.so
    ```

Embedded/setup.in.py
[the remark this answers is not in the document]
igalink: Thus, the definition of this variable must also be added to omnisci/scripts?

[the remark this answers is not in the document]
pearu: No, because `omniscidb/scripts/conda` scripts use the same build directory for all components, hence `build/Embedded/libDBEngine.so` exists when running `setup.py` 

The above does not apply to omniscidb-feedstock where `omniscidbe` and `pyomniscidbe` packages are built separately.

OSDependent/Unix/omnisci_path.cpp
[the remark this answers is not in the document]
pearu: Using `dlopen` flag `RTLD_NOLOAD` might resolve the problem.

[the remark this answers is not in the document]
igalink: Yes, actually this place determines whether the get_root_abs_path is located in one of the shared objects loaded by the calling application:
if (rc) {
...
} else {
      path_len = readlink("/proc/self/exe", abs_exe_path, sizeof(abs_exe_path));
}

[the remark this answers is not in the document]
igalink: > Using `dlopen` flag `RTLD_NOLOAD` might resolve the problem.

Maybe you're right. If the path generated for the libDBEngine is correct for applications that do not use this library (omnisci_server, initdb).

Embedded/setup.in.py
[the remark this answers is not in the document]
igalink: But in setup.py, without any conditions, you take a non-existent variable, build a path relative to it and add it to library_dirs. Are you sure there are no issues?

Embedded/setup.in.py
igalink: Thank you. Looks good

Embedded/setup.in.py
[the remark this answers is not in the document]
pearu: Fixed: `OMNISCI_ROOT_PATH` is used only when defined.

igalink (approved):

OSDependent/Unix/omnisci_path.cpp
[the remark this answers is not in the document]
pearu: Now using RTLD_NOLOAD dlopen: when handler is NULL, fallback to `/proc/self/exe` path to determine the omnisci root path.

Garra1980:
Just to check my understanding - say, I'm a developer and want to test fresh DBEngine functionality. what would be a proper way to install modified omniscidb*/pyomniscidb* components?

pearu:
> Just to check my understanding - say, I'm a developer and want to test fresh DBEngine functionality. what would be a proper way to install modified omniscidb*/pyomniscidb* components?

To install cpu-only omniscidbe packages, use
```
conda install -c conda-forge pyomniscidbe
```
To install cuda-enabled omniscidbe packages, use
```
conda install -c conda-forge "pyomniscidbe=*=*cuda"
```

Garra1980:
> > Just to check my understanding - say, I'm a developer and want to test fresh DBEngine functionality. what would be a proper way to install modified omniscidb*/pyomniscidb* components?
> 
> To install cpu-only omniscidbe packages, use
> 
> ```
> conda install -c conda-forge pyomniscidbe
> ```
> 
> To install cuda-enabled omniscidbe packages, use
> 
> ```
> conda install -c conda-forge "pyomniscidbe=*=*cuda"
> ```

No, I meant sort of local install, without conda-forge. If I want to test packages locally with some immediate changes.

pearu:
> No, I meant sort of local install, without conda-forge. If I want to test packages locally with some immediate changes.

There are many approaches possible to develop omniscidb from conda environment.

To develop omniscidb, I use conda environment created from

  https://github.com/Quansight/pearu-sandbox/blob/master/conda-envs/omniscidb-dev.yaml

that is activated using the following script:

https://github.com/Quansight/pearu-sandbox/blob/master/working-envs/activate-omniscidb-internal-dev.sh

that sets up some environment variables and displays basic development instructions.

Garra1980:
The thing is that there are already some scripts in master branch and we successfully use those scripts for development - see https://github.com/omnisci/omniscidb/blob/master/scripts/conda/build-install-all.sh. Not sure they will work after this PR though. 
And https://github.com/omnisci/omniscidb/blob/master/scripts/mapd-deps-conda-dev-env.yml helps build conda env for development (not quite correct I think)
Anyway I think we should provide (and document somewhere) the clear way of omnisci developing using conda env. Probably using scripts that are already committed. What do you think?

pearu:
> The thing is that there are already some scripts in master branch and we successfully use those scripts for development - see https://github.com/omnisci/omniscidb/blob/master/scripts/conda/build-install-all.sh. Not sure they will work after this PR though.

I don't see why the `build-install-all.sh` would not work with this PR. If it breaks something, it should be fixed.

> And https://github.com/omnisci/omniscidb/blob/master/scripts/mapd-deps-conda-dev-env.yml helps build conda env for development (not quite correct I think)
> Anyway I think we should provide (and document somewhere) the clear way of omnisci developing using conda env. Probably using scripts that are already committed. What do you think?

I think that `mapd-deps-conda-dev-env.yml` and `build-install-all.sh` is just another equivalent way to develop omniscidb from conda environment. In the end, I think it does not matter which instruction sets to use, however, my experience is that the instructions need to be maintained and kept up-to-date at all times by developers using these actively.

I use my yaml and development scripts mainly because keeping them up-to-date is efficient and matches my work style. For instance, I use [direnv](https://direnv.net/) a lot, keep the development system minimal, and use different environments for different tasks/projects, etc, all that to facilitate working with many projects at the same time. So, many details in my scripts might be irrelevant for other developers and could be simplified.

That said, I agree that it would be nice to have a document describing the basic development workflow to ease jumping into the omniscidb development process.

Garra1980:
> > The thing is that there are already some scripts in master branch and we successfully use those scripts for development - see https://github.com/omnisci/omniscidb/blob/master/scripts/conda/build-install-all.sh. Not sure they will work after this PR though.
> 
> I don't see why the `build-install-all.sh` would not work with this PR. If it breaks something, it should be fixed.

Right, right. I just meant we haven't tried it yet. Will check and fix in case of any issues

> 
> > And https://github.com/omnisci/omniscidb/blob/master/scripts/mapd-deps-conda-dev-env.yml helps build conda env for development (not quite correct I think)
> > Anyway I think we should provide (and document somewhere) the clear way of omnisci developing using conda env. Probably using scripts that are already committed. What do you think?
> 
> I think that `mapd-deps-conda-dev-env.yml` and `build-install-all.sh` is just another equivalent way to develop omniscidb from conda environment. In the end, I think it does not matter which instruction sets to use, however, my experience is that the instructions need to be maintained and kept up-to-date at all times by developers using these actively.
> 
> I use my yaml and development scripts mainly because keeping them up-to-date is efficient and matches my work style. For instance, I use [direnv](https://direnv.net/) a lot, keep the development system minimal, and use different environments for different tasks/projects, etc, all that to facilitate working with many projects at the same time. So, many details in my scripts might be irrelevant for other developers and could be simplified.
> 
> That said, I agree that it would be nice to have a document describing the basic development workflow to ease jumping into the omniscidb development process.

We will start handle this, I think

PR #649, opened 2021-04-22
Add DBE Python tests for RA and readCVS

igalink:
DBE Python tests  added for read CSV and execute a json that contains relational algebra instructions.


Garra1980:
Looks like this is in master and can be closed

PR #655, opened 2021-05-14
Gpu skeleton

kurapov-peter:


QueryEngine/NativeCodegen.cpp
```diff
@@ -2875,6 +3030,29 @@ Executor::compileWorkUnit(const std::vector<InputTableInfo>& query_infos,
   }
 
   // Generate final native code from the LLVM IR.
+  std::shared_ptr<CompilationContext> compilation_context;
```
fexolm, line 3033: It looks like this variable is unused. Maybe there is an issue here.

QueryEngine/NativeCodegen.cpp
```diff
@@ -2875,6 +3030,29 @@ Executor::compileWorkUnit(const std::vector<InputTableInfo>& query_infos,
...
+          query_func, multifrag_query_func, live_funcs, false, l0_mgr, co);
+      break;
+
+    default:
+      LOG(FATAL) << "Invalid device type";
+      return {};
+  }
   return std::make_tuple(
       CompilationResult{
           co.device_type == ExecutorDeviceType::CPU
```
kurapov-peter, line 3058: Looks like a merge bug, optimizeAndCodegen is called twice

QueryEngine/NativeCodegen.cpp
fexolm: Fixed

QueryEngine/NativeCodegen.cpp
fexolm: Fixed

PR #660, opened 2021-06-18
Fix race in TbbThreadPool.

ienkovich:
We might write to the `TbbThreadPool::results_` vector and try to extend it at the same time from different threads. Fix it by using a concurrent vector.


ienkovich:
Not needed for the new threading layer

PR #673, opened 2021-08-20
Introducing shim threading layer with TBB and standard library implementations

anton-malakhov:
Design choices are based on decisions and agreements between @ienkovich and @alexbaden


alexbaden:
Merged w/ 9c2bb95907ca03dc565f55f922d5343b8ff80371

PR #683, opened 2021-09-27
Spirv translator build script

kurapov-peter:
This builds spirv translator with a prebuilt llvm, worked for me with llvm 12, though I've seen many sanity tests failures with llvm 12.0.0


kurapov-peter:
I also had to modify the build script for it to successfully compile llvm12 (hence the second commit)

PR #695, opened 2021-11-03
Implement data removal for dropped FSI tables.

ienkovich:
Currently, when FSI table is dropped its data still lives. This creates memory leaks when OmniSci is used as a Modin backend. This patch implements data removal for dropped tables stored in FSI.


Garra1980:
Planned for 5.9

alexbaden:
Merged w/ 3ed20116a1f1ff0b06829043d7f6845d7c768882

PR #696, opened 2021-11-03
Remove unnecessary memory accesses from CPU codegen

de1iza:
It speeds up lauchCpuCode stage by around 30% for the following query for row-wise output: 
```sql
SELECT x*2 FROM table;
```

### DIFF
--- a/QueryEngine/GroupByAndAggregate.cpp
+++ b/QueryEngine/GroupByAndAggregate.cpp
@@ -1022,12 +1022,10 @@ llvm::Value* GroupByAndAggregate::codegenOutputSlot(
          null_key_lv,
          order_entry_lv});
   } else {
-    const auto output_buffer_entry_count_lv =
-        LL_BUILDER.CreateLoad(get_arg_by_name(ROW_FUNC, "max_matched"));
     const auto group_expr_lv =
         LL_BUILDER.CreateLoad(get_arg_by_name(ROW_FUNC, "old_total_matched"));
     std::vector<llvm::Value*> args{groups_buffer,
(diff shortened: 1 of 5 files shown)


ienkovich:
I guess splitting functional and non-functional changes (more accurate IR metadata) would simplify the review process.

QueryEngine/ir.ll
```diff
@@ -0,0 +1,254 @@
+
```
fexolm, line 1: Please remove this file

QueryEngine/GroupByAndAggregate.cpp
```diff
@@ -868,58 +870,64 @@ bool GroupByAndAggregate::codegen(llvm::Value* filter_result,
...
-          if (query_mem_desc.didOutputColumnar()) {
-            nullcheck_cond = LL_BUILDER.CreateICmpSGE(std::get<1>(agg_out_ptr_w_idx),
-                                                      LL_INT(int32_t(0)));
-          } else {
-            nullcheck_cond = LL_BUILDER.CreateICmpNE(
-                std::get<0>(agg_out_ptr_w_idx),
-                llvm::ConstantPointerNull::get(
-                    llvm::PointerType::get(get_int_type(64, LL_CONTEXT), 0)));
+      {
+        AUTOMATIC_IR_METADATA(executor_->cgen_state_.get());
```
fexolm, line 874: Please remove this indent and all metadata changes as it is irrelevant

corecursion:
Interesting. I'm curious about how long the typical launch time is that you've seen. (I assume the launch time is where the 30% speedup is.)

PR #704, opened 2021-11-15
Optimize count distinct reduction time

fexolm:


QueryEngine/HyperLogLog.h
```diff
@@ -25,6 +25,7 @@
 #ifndef QUERYENGINE_HYPERLOGLOG_H
 #define QUERYENGINE_HYPERLOGLOG_H
 
+#include <tbb/parallel_for.h>
```
fexolm, line 28: use threading layer

PR #719, opened 2022-01-17
fix bugs for llvm patch file

sijialouintel:
Omnisci use this file to patch llvm 9 for glibc 2.31+ support. The original patch file can't be parsed successfully and are considered as garbage. This updated patch is generated by command 'git diff', which can be used by
`sudo patch -p0 < $SCRIPTS_DIR/llvm-9-glibc-2.31-708430.patch`
(line 193 in https://github.com/omnisci/omniscidb/blob/master/scripts/common-functions.sh)


simoneves:
Closing as obsolete

PR #763, opened 2022-11-10
Assign a request_id to each thrift call and optionally log parent/child

jack-mapd:
Signed-off-by: jack <jack@heavy.ai>

### DIFF
--- a/ThriftHandler/DBHandler.cpp
+++ b/ThriftHandler/DBHandler.cpp
@@ -64,6 +64,7 @@
 #include "QueryEngine/TableFunctions/TableFunctionsFactory.h"
 #include "QueryEngine/TableOptimizer.h"
 #include "QueryEngine/ThriftSerializers.h"
+#include "RequestInfo.h"
 #include "Shared/ArrowUtil.h"
 #include "Shared/DateTimeParser.h"
 #include "Shared/StringTransform.h"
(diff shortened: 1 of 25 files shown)


ThriftHandler/DBHandler.cpp
```diff
@@ -491,10 +499,11 @@ void DBHandler::krb5_connect(TKrb5Session& session,
   THROW_DB_EXCEPTION("Unauthrorized Access. Kerberos login not supported");
 }
 
-void DBHandler::connect(TSessionId& session,
+void DBHandler::connect(TSessionId& session_id,
                         const std::string& username,
                         const std::string& passwd,
                         const std::string& dbname) {
+  logger::set_new_request_id();
   auto stdlog = STDLOG();  // session_info set by connect_impl()
```
mattpulver, line 507: Comment in `internal` is `// session set by connect_impl()`.
(Should probably be `session_id` but we should fix it upstream first.)

mattpulver (commented):
To simplify this, I reviewed the patch generated from:
```
diff -Naur ThriftHandler/DBHandler.cpp heavydb-internal/ThriftHandler/DBHandler.cpp
```

Looks like these two methods are missing the upstream changes:
 * `DBHandler::get_session_info()`
 * `DBHandler::clear_cpu_memory()`

PR #770, opened 2022-12-13
Update README post-rebrand

tmostak:
Update doc and other links to heavy.ai, process names to `heavydb`, `heavysql`, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,12 @@ HEAVY.AI provides pre-built binaries for Linux for stable releases of the projec
 
 | Distro | Package type | CPU/GPU | Repository | Docs |
 | --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


cdessanti (approved):
The changes are ok for me.

we should just be using uniform policies for the releases on centos and ubuntu
https://releases.heavy.ai/os/yum/stable/(cpu|cuda) are visible while
https://releases.heavy.ai/os/apt/stable/(cpu|cuda) are not (we have other links like https://releases.heavy.ai/os/apt/pool/(cpu|cuda)/h/heavyai/ that list the packages)

also the oxygen it's generated with omnisci brand, but I think this is out of the scope of the PR)

jack-mapd (commented):
The changes look fine.  Clicked on most of the links and didn't get any errors.

I also ran a cmake with all of the options specified on the README and MAPD_DOCS_DOWNLOAD isn't around any more - not sure what it was replaced by. A quick grep of the cmake file didn't show any documentation style flags.  Also PREFER_STATIC only works on Centos not on Ubuntu which should be mentioned.

PR #774, opened 2023-01-05
Update system dynamically when new license is set.

jack-mapd:
This PR fixes a few issues:

Fixes persistence of num sessions.
Currently there is no distinction between the number of sessions specified by the command line and the number of sessions persisted between restarts, so we had no way of telling if the restarted value came from a cmd line parameter or a previous license and which should take priority. This is fixed by creating a new record of the number of sessions in the DBhandler so that we have a distinction between the current value, the cmd-line value, and the license value.

Signed-off-by: jack <jack@omnisci.com>


ThriftHandler/DBHandler.cpp
```diff
@@ -492,6 +502,43 @@ bool DBHandler::isAggregator() const {
   return leaf_aggregator_.leafCount() > 0;
 }
 
+<<<<<<< HEAD
```
misiugodfrey, line 505: Remove diff-generated lines.

misiugodfrey (approved):

PR #787, opened 2023-06-27
Misiu/os advance

misiugodfrey:
Catch OS up to 7.0 deps


misiugodfrey:
Part of the updates bumped clang format to use clang version 14 which changed some of the rules.  The automated merge run doesn't seem to be using the new dependency, so that's what's causing the discrepancy.

PR #843, opened 2025-02-20
Update QueryExporterGDAL.cpp

mflaxman10:
Change export type of kDATE in a shape file to OFTDate in OGR/GDAL types.

From docs:
"In GDAL OGR, the type used for dates is OFTDate. This corresponds to a date field that stores only the year, month, and day, without any time or timezone information."


mflaxman10:
Requesting this update (and any related test..) to support "date" column being exported as a shape file date column instead of as a string.   Simple date columns are common in oil and gas, and so useful to export.   ISO datetimes by contrast are not supported by shape files and current method of exporting those to TEXT is best choice available since it avoids loosing the time component information.

simoneves:
Fixed as part of SIO-1878 (2/27/25)

PR #846, opened 2025-05-26
Misiu/fix merge breaks

misiugodfrey:
This PR gets OS up to 9.0.  Builds and tests successfully.


tmostak:
@misiugodfrey this is great. Do we need to change anything in the readme to reflect any changes to the build or dep install process?

simoneves:
Consequently, `MAPD_RENDERING_LIBRARIES`, `ENABLE_RENDER_TESTS`, `ENABLE_RENDER_LOGGER`, and `ENABLE_SHADER_DEBUG_PRINTF`.

simoneves (approved):
I have not reviewed the code changes, as there are over 1000 touched files.

I propose a revision to the OS conversion process by which we also strip EE-only options such as `ENABLE_RENDERING` (and all the consequential logic), so that they're not even selectable, to avoid confusion or expectation by the uninformed. Also strip some secondary CMake logic to do with AWS Licensing and `OS_SPECIFIC_LICENSING_LIBS`.

Given that all that stuff has always still been in OS, I'm not gonna hold this release up for it. I was just surprised that it was there.

Otherwise, tested with shared deps on Ubuntu 24, builds clean, and runs interactively fine with simple tests. Running full tests now. Will come back and revert my approval if anything fails! :)

simoneves (changes requested):
`ThirdParty/gdal-data` needs removing

simoneves:
'ThirdParty/stb` and `ThirdParty/egl` are probably also superfluous.

misiugodfrey:
> I propose a revision to the OS conversion process by which we also strip EE-only options such as `ENABLE_RENDERING` (and all the consequential logic), so that they're not even selectable, to avoid confusion or expectation by the uninformed. Also strip some secondary CMake logic to do with AWS Licensing and `OS_SPECIFIC_LICENSING_LIBS`.

I think this is a good idea.  Rather than strip this code out of the CMake files directly, I would suggest we add this checking to the the internal version of the CMake files and then propagate it to OS, that way the files stay in sync and avoid future merge conflicts.  If we are board for that direction, then I don't think we need to wait for such a change to merge this PR, as it would be a QOL improvement for down the line.

mattgara (changes requested):
I have some feedback but can not write it inline as GitHub is impossible to use to leave comments inline with the code with this large a PR (at least for me.)

ThirdParty/gdal-data/esri_Wisconsin_extra.wkt
```diff
@@ -1,144 +0,0 @@
-103300,PROJCS["NAD_1983_HARN_WISCRS_Adams_County_Meters",GEOGCS["GCS_North_American_1983_HARN",DATUM["D_North_American_1983_HARN",SPHEROID["GRS_1980",6378137.0,298.257222101]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Transverse_Mercator"],PARAMETER["False_Easting",147218.6942],PARAMETER["False_Northing",0.0037],PARAMETER["Central_Meridian",-90.0],PARAMETER["Scale_Factor",1.0000365285],PARAMETER["Latitude_Of_Origin",43.36666666666667],UNIT["Meter",1.0]]
```
simoneves, line 1: This was the only file? Even weirder! But correct to remove it

docker/ee/LicenseServer/Dockerfile
paul-aiyedun, line 1: Remove the `ee` directory.

scripts/ee/start_heavyiq.sh
paul-aiyedun, line 1: Remove the `ee` directory.

ThriftHandler/CommandLineOptions.cpp
```diff
@@ -121,6 +139,11 @@ void CommandLineOptions::fillOptions() {
                          ->default_value(g_bigint_count)
                          ->implicit_value(true),
                      "Use 64-bit count.");
+  desc.add_options()("llm-transform-max-num-unique-value",
```
paul-aiyedun, line 142: This only applies to EE.

paul-aiyedun (changes requested):

mattgara:
Also these file paths should probably simply be removed with all contents:

```
./docker/ee
./docker/ee/LicenseServer
./docker/ee/LicenseServer/Dockerfile
./scripts/ee
./scripts/ee/start_heavyiq.sh
```

andrewseidl:
The headers in the `/os/` subdirs are minimal stubs for the EE equivalents - they're not actually used on the OS side of things, but we took the stub approach way back in the day in order to minimize the number of `#ifdef`s in the internal codebase (which also makes it easier to pull code in from the OS repo).

A quick grep showed that they're mostly still needed (eg `PkiServer.h` has `PkiServer` -> `pki_server_`, which is used in `SysCatalog.cpp`).

paul-aiyedun (approved):

simoneves (approved):
LGTM. What do we need to do to satisfy the CLA thing?

andrewseidl:
> LGTM. What do we need to do to satisfy the CLA thing?

Should be able to ignore it, I don't think it was ever set as blocking merge. Ideally someone with repo admin privs could take a look at the webhooks and settings to see if it can be removed. Though I want to say the last time I looked at it, one of the rebrands broke something on cla-assistant's side of things, so you might have to get in contact with them to fix/remove it.

mattgara (approved):
LGTM

misiugodfrey:
Apparently the Github UI has a limit of 100 commits in a rebase&merge operation, so I have had to break this PR down into 4 parts and merge them separately.  They have now been merged successfully.

